PR #174, opened 2017-11-30
Add a connector for slack

LTMXcitrus:
Hello there! 

Here is a proposition for a slack connector.
It's quite basic for now, but I think it's nice and I use it myself :)


vsct-jburet:
Thanks a lot!  :) I merge the PR

PR #218, opened 2018-03-07
NLP API - add getIntentByNamespaceAndName service

elebescond:
Add NLP API route /rest/nlp/intents?namespace=vsc&name=test


vsct-jburet:
Thanks! - merged :)

PR #255, opened 2018-05-04
trigger word are not mandatory anymore and the bot can't answer bots messages

Bybeez:
Changed the trigger_word type to a nullable to make the bot able to receive messages without trigger_word.
Added a condition so the bot can't answer a message from a bot.


vsct-jburet:
Thanks

PR #302, opened 2018-07-19
Fix potential race conditions in mongo-based user lock implementation

Bertrand:
Current implementation of user lock is not atomic and thus prone to race conditions.

Proposed implementation fixes the issue by using an atomic findAndModify (with upsert).


vsct-jburet:
Thanks. included in 2.0.1-SNAPSHOT :)

PR #418, opened 2019-03-26
Proxy configuration to HTTP client

FabienPennec:
Add the possibility to configure a http proxy to Retrofit.


vsct-jburet:
We use the java system properties ( https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html ) on our side, but this is an other option. I merge the PR.

PR #424, opened 2019-03-28
truncate label and description quick replies for twitter connector

auroredea:
Truncate to not fail into some unwanted errors.


vsct-jburet:
merged

PR #427, opened 2019-04-01
Alexa Connector - Fix intent name with suffix

FabienPennec:
Fix incorrect substring on the intent name in AlexaTockMapper.


vsct-jburet:
Thanks

PR #438, opened 2019-04-16
add business chat connector

ReyAnthony:
The implementation of the CSPBusinessChatClient is to be made on the bot side according to the model chosen by the CSP.


vsct-jburet:
merged

PR #451, opened 2019-04-25
Implementation of businessChat List picker

mrboizo:
In the scope of the new business chat connector, we have implemented the support of the list picker template


vsct-jburet:
merged

PR #455, opened 2019-04-29
some fixes for business chat

ReyAnthony:
Made some modifications so that the connector will not break if there is an unhandlded MessageType (example : alcmeon handover has a specific type which we didnot handle and don't want to expose)


ReyAnthony:
will take another approach

PR #462, opened 2019-05-07
Visibility modifications to allow an implementation of NlpController in a custom module

FabienPennec:
I'm currently coding a new implementation of the interface NlpController in a custom module.
Unfortunately, I can't access some properties or constructors because their visibility is internal to the module tock-bot-engine.

This pull request is to allow an implementation of NlpController outside tock-bot-engine.


vsct-jburet:
Merged. I just have reverted the public visibility of BotRepository.nlpListeners. I prefer to add a BotRepository.forEachNlpListener (see https://github.com/voyages-sncf-technologies/tock/commit/30235e9e2e657450f7b79dbdd3ef808e60c0dac7)

PR #491, opened 2019-06-11
Add method to disable bot outside the TockBotBus

elebescond:


elebescond:
Actually not really useful

PR #502, opened 2019-06-19
Add a parameter to handle extended user profile fields on facebook

bakic:


bakic:
A new parameter _tock_messenger_extended_profile_fields_ was added to handle the fields that will be requested frm facebook concerning the user profile.
By default the parameter is set to **false** and the fields _first_name, last_name, profile_pic_ are requested from facebook
If special authorizations were set, the paramete rcan be set to true to get _locale, timezone, gender fields_
More info in: https://developers.facebook.com/docs/messenger-platform/identity/user-profile/#fields

PR #522, opened 2019-07-16
Add an intent importer from Dialogflow to Tock to simplify the migration process

FabienPennec:
Useful to reuse a FAQ Dialogflow project and import it in Tock.


vsct-jburet:
Thanks, merged.

PR #684, opened 2019-11-21
669 add information xray test execution

kchiron:


bot/admin/test/xray/src/main/kotlin/XrayService.kt
```diff
@@ -582,15 +584,75 @@ class XrayService(
...
+                }
+
+        return XrayTest(jira.key)
+    }
+
+    fun updateXrayTest(dialog: DialogReport, testKey: String): XrayTest? {
+        val steps = defineTestSteps(dialog)
+
+        if (XrayClient.getIssue(testKey) == null) {
+            return null
```
curinator: it seems this return is unreachable because `getIssue` returns `error` if we are unable to get the issue

bot/admin/test/xray/src/main/kotlin/XrayClient.kt
```diff
@@ -179,6 +183,11 @@ object XrayClient {
 
     fun saveStep(testKey: String, step: XrayBuildTestStep) = xray.saveStep(testKey, step).execute().body()
 
+    fun deleteStep(testKey: String, stepId: Int) =
```
curinator: IMHO `Long` to `Int` conversion should be in `xray.deleteStep`call

bot/admin/test/xray/src/main/kotlin/XrayService.kt
```diff
@@ -582,15 +584,75 @@ class XrayService(
...
+        }
+
+        // remove all existing test steps
+        val stepList = XrayClient.getTestSteps(testKey)
+        stepList.forEach { currentStep ->
+            XrayClient.deleteStep(testKey, currentStep.id.toInt())
```
curinator: What about remaining tests if one deletion fails ? Do we assume others will fail too ?

bot/admin/test/xray/src/main/kotlin/XrayService.kt
```diff
@@ -582,15 +584,75 @@ class XrayService(
...
+                            XrayClient.addTestToTestPlan(jira.key, labelTestPlansMap[it]!!)
+                        }
+            }
+        }
+
+        steps?.forEach {
+            XrayClient.saveStep(jira.key, it)
+        }
+
+        XrayPrecondition.getPreconditionForUserInterface(dialog.actions.first().userInterfaceType)
```
curinator: Are we sure that `actions` will never be empty at this stage  ? If not `first()` failure should be handled

bot/admin/test/xray/src/main/kotlin/XrayTestService.kt
```diff
@@ -18,8 +18,10 @@ package ai.tock.bot.admin.test.xray
 
 import ai.tock.bot.admin.test.TestCoreService
 import ai.tock.bot.admin.test.TestService
+import ai.tock.bot.admin.test.model.BotDialogRequest
```
curinator: unused imports

bot/admin/test/core/src/main/kotlin/TestPlanService.kt
```diff
@@ -144,8 +146,12 @@ object TestPlanService {
             }
         }
 
-        // save the test plan execution into the database
-        testPlanDAO.saveTestExecution(exec.copy(status = TestPlanExecutionStatus.COMPLETE))
+        if (nbErrors != 0) {
+            exec.nbErrors = nbErrors
```
curinator: It seems odd to modify `nbErrors` to `var` only for this case. Why don't you return the copy below which has the right error count and keep `nbErrors` as `val` ?

bot/admin/test/xray/src/main/kotlin/XrayClient.kt
```diff
@@ -45,6 +45,7 @@ import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.MultipartBody
 import okhttp3.RequestBody
 import okhttp3.ResponseBody
+import org.jetbrains.annotations.TestOnly
```
curinator: unused import

bot/admin/test/xray/src/main/kotlin/XrayClient.kt
```diff
@@ -135,14 +136,17 @@ object XrayClient {
 
     fun getProjectFromIssue(issueKey: String): JiraTestProject {
         val klaxon = Klaxon()
+        val xrayResponse = xray.getIssue(issueKey).execute()
+
+        return if (!xrayResponse.isSuccessful) {
```
curinator: IMHO it would be more readable to handle successful case first and avoid negation here

bot/admin/test/xray/src/main/kotlin/XrayService.kt
kchiron: I've removed the nullity test but left the getIssue() call to be able to make the difference between "issue does not exist" or "no steps available for the current test" errors.

bot/admin/test/xray/src/main/kotlin/XrayClient.kt
kchiron: Moved

bot/admin/test/xray/src/main/kotlin/XrayClient.kt
kchiron: right :)

bot/admin/test/xray/src/main/kotlin/XrayClient.kt
kchiron: You're right, I've reversed the condition. Now successful case is handled first.

bot/admin/test/core/src/main/kotlin/TestPlanService.kt
kchiron: Indeed, now only "copy()" is used to update value in MongnDB. On your advice, I've also store the exec copy into another variable to return the right updated one.

bot/admin/test/xray/src/main/kotlin/XrayTestService.kt
kchiron: Removed

bot/admin/test/xray/src/main/kotlin/XrayService.kt
kchiron: I don't know much about XrayPrecondition. I just extracted this step definition from generateXray() because this part would be used bot in generateXray() and updateXrayTest() methods.
I think that dialog would be filled at least with a greeting but I'm not sure. @vsct-jburet may answer better than me to this :).

bot/admin/test/xray/src/main/kotlin/XrayService.kt
kchiron: If steps deletion fails, en error would be raised and test update will be called again by testers. So, updateXrayTest() will be called again and it will try to delete steps.
No error handler will be added here.

bot/admin/test/xray/src/main/kotlin/XrayService.kt
kchiron: After discussing with @vsct-jburet , dialog contains at least a greeting, then dialog is never empty.

PR #699, opened 2019-12-04
Swagger Web Connector

dabadiesimon:
Adding Swagger for the Web Connector based on the existing description.


vsct-jburet:
Merged. Thanks.

PR #745, opened 2020-01-17
Add a MongoDB credentials provider

FabienPennec:
Add a MongoDB credentials provider to provide credentials from another source than the MongoDB connection url.
The provider is called only if there is no credentials in the Mongo connection url.

By default, the provider provides no credentials.


vsct-jburet:
merged. Thanks.

PR #754, opened 2020-01-21
fixes #735 Prevent spaces in path

### LINKED ISSUE #735: Connector  path should not contain spaces
In the Tock Studio "Configuration" view, entering a Connector Name with spaces results in Connector Identifier and Relative REST Path containing spaces, which creates errors (eg. testing the bot from the Test view returns "technical error").

Tock Studio should prevent users from creating Connector IDs and paths with special chars, at least space chars. Possible solutions : replace spaces automatically and/or to add form field validation.

francoisno:
I replaced spaces by underscores '_' where applicationId is used in : 
- Connector path
- REST connector "test" path

Did not replace other special chars.
Did not replace spaces, where namespace is used for path (but namespace cannot include space right?).

### DIFF
--- a/bot/admin/web/src/app/configuration/bot/new-bot.component.ts
+++ b/bot/admin/web/src/app/configuration/bot/new-bot.component.ts
@@ -65,7 +65,7 @@ export class NewBotComponent implements OnInit {
   validate() {
     const locale = this.firstFormGroup.value.firstCtrl;
     const channel = this.channel.value;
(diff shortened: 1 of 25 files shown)


vsct-jburet:
@francoisno  I think it would be better to remove space when creating/updating connectorId

francoisno:
New commit : as discussed, I have added a 'label' property to applications.
When creating an application, its name is the label without special chars by default. Name is still editable.
For previously-created applications (in base or dumps without label), label is copied from the name by default.

vsct-jburet:
Merged. Great ! :)

PR #757, opened 2020-01-21
fixes #731 trigger targetIntent after answer

### LINKED ISSUE #731: Target intent not triggered when the optionnal answer is specified in an action
When building a story in the admin UI:
If an action is added with a non-empty answer and a target intent is specified, the answer will be displayed but the story associated with target intent will not be triggered.

72157670-c5938580-33b8-11ea-98b6-d78d5602f692

francoisno:


vsct-jburet:
Merged

PR #773, opened 2020-02-11
Improve detailed heathcheck

kornfleks:
Expose database name constants in shared.
Refracto of `pingMongoDatabase` in `Mongos.kt`.


vsct-jburet:
Thank you

PR #778, opened 2020-02-14
785 fixes search filter disabled after story update

francoisno:
Fixes issue #785 (moved from former repository).

I suppose it is not necessary to reset state after a story update, considering intent modifications (eg. modifying the intent or its category) are performed in an additional dialog with its own Update button resetting state.


vsct-jburet:
@francoisno no, resetState is mandatory. What is the issue? the old project is removed now :(

francoisno:
Ok, since resetState is mandatory there, I guess one should subscribe somehow to this event, in order to empty filters / or preferably keep filters and perform search again.

PR #794, opened 2020-02-24
[messenger connector] button template constraints

auroredea:
Button template on messenger connector constraints to text limit (640 characters) and numbers of button (1 to 3)


vsct-jburet:
merged

PR #920, opened 2020-05-19
fixes #919 Duplicate step label errors

### LINKED ISSUE #919: Duplicate step errors and text not duplicated correctly
Duplicate step labels is throwing errors when subtitle is undefined or null and text is not duplicated correctly.

broxmik:


vsct-jburet:
merged

PR #964, opened 2020-06-16
resolves #849 user namespaces: add a way to create new namespace

### LINKED ISSUE #849: user namespaces: add a way to create new namespace
Today when you create a namespace without an app in the db, there is an error

vsct-jburet:
Please review this PR, especially the web interface updates :

-nebula components
- focus tip, (keyup.enter) & tooltip


mrboizo:
Review returns : 
- responsive composant looks OK
- feature works
- the Travis is KO
- choice of Nebular instead of Material ?
- what about the suppression of a namespace ?

PR #973, opened 2020-06-25
resolves #963 improved i18n import and export components

### LINKED ISSUE #963: Add action that exports only the filtered i18n values
There is a need to export / import only filtered i18n labels.

broxmik:


vsct-jburet:
A bit more kdoc/test would be nice & also don't forget do set in your IDE preference to never made * import.

Very nice, thanks :)

PR #985, opened 2020-07-02
942 analytics + 850 menu/tabs

francoisno:
No database change at this point, using only previous stats.

Developed and tested with @bakic on demo platform and Marvin database.
To be tested with Jarvis data for possible perf issues.


francoisno:
@bakic Travis fails with `[ERROR] /home/travis/build/theopenconversationkit/tock/bot/storage-mongo/src/main/kotlin/UserTimelineCol.kt:[125,23] Unresolved reference: ofInstant`

PR #1000, opened 2020-07-10
fixes #999: Allow to use named argument in format message label

### LINKED ISSUE #999: [Enhancement] Allow to use named argument in i18n sentences pattern

curinator:
Allow user to use named arguments in i18n labels.

- allowed characters : a-z, A-Z and underscore
- argument name format {:name}

Signed-off-by: Toki RAOSETA <toki.raoseta@gmail.com>

### DIFF
--- a/translator/core/src/main/kotlin/NamedArgumentNormalizer.kt
+++ b/translator/core/src/main/kotlin/NamedArgumentNormalizer.kt
@@ -1,0 +1,48 @@
+package ai.tock.translator
+
+import mu.KotlinLogging
+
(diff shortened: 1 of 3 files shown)


curinator:
Limitation: Mix annotation is supported by `NamedArgumentNormalizer`but **not supported** by its integration due to `escapeQuotes(NamedArgumentNormalizer.normalize(text).label)` which doesn't provide label arguments. Without arguments, we are not able to detect easily vanilla arguments count which prevent us to generate enriched argument indexes.

PR #1103, opened 2020-11-03
fixes #1053 redirection view is broken when a story is missing

### LINKED ISSUE #1053: Redirection Story Rules view is broken when a story is missing
- Create a new Redirection Story Rule ( story1 -> story2)
- Delete story2
- Redirection Story Rules view is broken

vsct-jburet:


broxmik (approved):

francoisno (approved):
@vsct-jburet  how is it possible to have no `switchToStory` since the feature in `redirectedFeatures = features.filter(feat => feat.switchToStoryId != null)`?
Is after modifying a story rule? Rather perform a missing `refresh()` somewhere?
Or maybe `switchToStory` is not null but empty?

PR #1109, opened 2020-11-09
fixes #1070 Studio: improve encrypt error message

### LINKED ISSUE #1070: [Studio] Fix error or improve message when entity obfuscated w encryption not configured
Currently if no `tock_encrypt_pass` is configured, it is possible to set an entity type _Obfuscated_ in Tock Studio, yet going to _Inbox_ raises a technical error (toaster without details).
Looking at logs, it's a `no tock_encrypt_pass` exception.
Either fix it by having a default key, or hide the _Obfuscated_ option when not set, or improve the error message in both logs and UI.

francoisno:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,4 +39,4 @@ Every new feature or fix should embed its unit test(s).
 
 More about [sources and contrib](http://doc.tock.ai/tock/en/about/contribute/).
 
-Feel free to [contact us](contact.md).
+Feel free to [contact us](https://doc.tock.ai/tock/en/about/contact/).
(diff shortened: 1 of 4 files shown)


vsct-jburet:
@francoisno I would prefer to use a dedicated exception instead of testing the exception message... :)

francoisno:
> @francoisno I would prefer to use a dedicated exception instead of testing the exception message... :)

Amended with `NoEncryptionPassException` for a (slightly) more specific message:

```
                } catch (t: NoEncryptionPassException) {
                    logger.error(t)
                    badRequest("Error obfuscating sentences: ${t.message}")
                } catch (t: Exception) {
                    logger.error(t)
                    badRequest("Error searching sentences: ${t.message}")
                }
```

vsct-jburet (approved):

PR #1133, opened 2020-12-07
fixes #1075 Dark Mode issues in Intents

### LINKED ISSUE #1075: [Studio] Dark Mode issues in Intents
Intents view has dark-mode issues:

Capture d’écran du 2020-10-26 11-15-38

bakic:


francoisno (changes requested):
Thx for the PR.
Comparing with demo.tock.ai, I found these issues to be fixed:

- Table content disappears in background when dialogs are open (add shared intent / add state)
- Table/columns are visible when there is no intent
- Broken download-sentence-dump button (`downloadSentencesDump is not a function`)

bakic:
Concerning the first point, I found it like that and I noticed it but didn't want to touch it because I don't know why it was implemented like that. I can remove this behaviour if it's useless.

francoisno:
Just pushed fixes for the previously-noticed issues, yet `detectChanges` is a dirty hack, and `display` seems useless to me so... please review this! :wink:

bakic:
The changeDetector was added by @vsct-jburet when dealing with perf issues but I can't find why it was added before displaying dialogs.
I made a test without the changeDetector and it's working fine.
@vsct-jburet can you help us with this?

vsct-jburet:
> @vsct-jburet can you help us with this?

@bakic when opening a dialog from a page with many dom components, the opening was (very) slow. (see https://github.com/angular/components/issues/11330 ). I fixed it by removing the dom when opening the dialog and restoring the dom when closing the dialog.  To reproduce, just display a search with 100 sentences and open a dialog - or go to the intent page with many intents (https://demo.tock.ai/nlp/intents) and open a dialog. May be with the nebular version the issue is no longer present and you can remove the hack :)

bakic:
@francoisno I tested without the freeze before displaying the dialog on a list with more than 100 intents and it's fine.
I pushed the fix.
Can you please check if it's fine now, thanks

vsct-jburet (approved):

francoisno (approved):

PR #1149, opened 2021-01-06
Doc: 1147 slack guide + 1148 fix search (english)

francoisno:


djowalker (approved):
Everything seems to be releveant according to the issue.

PR #1151, opened 2021-01-07
fixes TinyMCE vulnerability

francoisno:
https://github.com/theopenconversationkit/tock/security/dependabot 
TinyMCE is only used for _Settings > Logs > View Full Log_, right ?


bakic:
To view the full logs in LOGS tab, jsoneditor library is used.
I can't find where are we using TinyMCE.
Maybe we should remove it

vsct-jburet:
@bakic can you try to remove it ? :)

bakic:
@vsct-jburet I removed the dependency with no impact noted

PR #1165, opened 2021-01-18
fixes #1164 Git: add .gitattributes to better Windows DX

### LINKED ISSUE #1164: [Git] Allow Windows developers to see correct git status
As a Windows developer, I didn't find configuration for line ending defined at repository level (no .gitattributes).
This leave each Window developper to configure his working station to prevent future conflicts with line ending.

Furthermore, after building tock (mvn install), multiples files are generated (\*\*/target/generated-sources/kapt/compile/\*\*).
It's kind of strange theses files are tracked by Git in the first place but it might be needed (for obscure reason).
Then, when I do a "git status", all these generated files appear as modified while only line ending changed.

To prevent this and globally all future line-ending problems, using a .gitattributes file would be a solution.

It allows to define a default checkout method et custom checkout method depending on certains files (cf https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#per-repository-settings).

A working example for the situation described above : 

```
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto

# For generated-sources
**/target/generated-sources/kapt/compile/** text eol=lf
```
For "generated-sources" ' special case, it might be simpler to untrack theses files if there are not really needed as tracked.

fbelliard:
fixes #1164 

Add .gitattributes to ensure in which fashion every file should be checkout, depending on developer's OS.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# For generated-sources, force LF to prevent diff showing for Windows users
+**/target/generated-sources/** text eol=lf


vsct-jburet (approved):

vsct-jburet:
Thank you

PR #1200, opened 2021-02-25
Studio: add tooltips to the channel icons in the Answers menu

Fabilin:
While learning this tool, I noticed the icons denoting the type of channel for which an answer is applicable but had to ask what they meant as their purpose was not immediately clear to me. I believe the application could get a small boost in usability from adding tooltips to those icons, so I made a PR to do exactly that.

location

Tooltip for text answers:
image

Tooltip for voice answers:
image

### DIFF
--- a/bot/admin/web/package-lock.json
+++ b/bot/admin/web/package-lock.json
@@ -15232,6 +15232,7 @@
           "dev": true,
           "optional": true,
           "requires": {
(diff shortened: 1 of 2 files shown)


francoisno (approved):

vsct-jburet (approved):

vsct-jburet:
Thank you

PR #1227, opened 2021-03-10
resolves #1226 UrlButton: add an option to open the url in the same o…

### LINKED ISSUE #1226: [WEB_CONNECTOR] UrlButton: add an option to open the url in the same or another window
The urlButton allows to send a button with a url.
For the moment the choice to open the url in the url in the same window or in another one is mode in the client side.
Adding a parameter to the UrlButton indicating the client how the url should be opened from the server side will give more flexibility to the component.

bakic:
…r another window

### DIFF
--- a/bot/connector-web/src/main/kotlin/WebBuilders.kt
+++ b/bot/connector-web/src/main/kotlin/WebBuilders.kt
@@ -102,12 +102,25 @@ fun <T : Bus<T>> T.webButton(
 fun <T : Bus<T>> T.webUrlButton(
     title: CharSequence,
     url: String,
-    imageUrl: String? = null
+    imageUrl: String? = null,
+    target: HrefTargetType = HrefTargetType._blank
+): Button =
+    webUrlButton(title, url, imageUrl, target.name)
+
+/**
+ * Creates a url button
+ */
+fun <T : Bus<T>> T.webUrlButton(
+    title: CharSequence,
+    url: String,
+    imageUrl: String? = null,
+    target: String? = null
 ): Button =
     UrlButton(
         translate(title).toString(),
         url,
-        imageUrl
+        imageUrl,
+        target
     )
 
 /**
--- a/bot/connector-web/src/test/resources/card_with_buttons.json
+++ b/bot/connector-web/src/test/resources/card_with_buttons.json
@@ -20,6 +20,7 @@
             "title": "title",
             "url": "http://www.sncf.com",
             "type": "web_url",
+            "target": "_blank",
             "clazz": "url_button"
           }
         ]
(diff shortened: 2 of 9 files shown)


vsct-jburet:
@bakic does it means that the legacy behaviour is target:blank for the url ?

bot/connector-web/src/main/kotlin/WebBuilders.kt
```diff
@@ -103,11 +103,13 @@ fun <T : Bus<T>> T.webButton(
 fun <T : Bus<T>> T.webUrlButton(
     title: CharSequence,
     url: String,
+    openInSameWindow: Boolean = false,
```
francoisno, line 106: Adding the new param after `imageUrl` wouldn't be safer? (for callers, not naming the params)

bot/connector-web-model/src/main/kotlin/ai/tock/bot/connector/web/send/UrlButton.kt
```diff
@@ -22,5 +22,6 @@ import com.fasterxml.jackson.annotation.JsonTypeName
 data class UrlButton(
     val title: String,
     val url: String,
+    val openInSameWindow: Boolean = false,
```
francoisno, line 25: Maybe `target` would be a better name than `window` and `targetBlank` or `targetSelf` better, to fit W3C attribute? (https://www.w3schools.com/tags/att_a_target.asp)

bot/connector-web/src/test/resources/card_with_buttons.json
```diff
@@ -20,6 +20,7 @@
             "title": "title",
             "url": "http://www.sncf.com",
             "type": "web_url",
+            "openInSameWindow": "false",
```
francoisno, line 23: `"openInSameWindow": false,` sans les quotes ? (pour être iso les autres fichiers de test)

francoisno (changes requested):

bot/connector-web/src/main/kotlin/WebBuilders.kt
bakic: ok

bot/connector-web-model/src/main/kotlin/ai/tock/bot/connector/web/send/UrlButton.kt
bakic: I changed the name of the attribute to target and added an enum that gathers the possible values of the target param to fit W3C values for the target attribute

bot/connector-web/src/test/resources/card_with_buttons.json
bakic: @francoisno please check the new commit

francoisno: I just pushed a fix of test resources (2 `openInSameWindow` remained).
Please let me know if it's ok for you @bakic

francoisno (approved):

PR #1237, opened 2021-03-21
resolves #1166 Core: render carousels from consecutive cards

### LINKED ISSUE #1166: Build a Carousel Story inside Tock Studio
Hi Tock Team,

i have a bot usage that is only made of fullText queries answers.
Some complex queries are answered with full text carrousels.

ask me if you need a sample.

So my question is: Is it possible to build carrousels inside a story.

If we can do this so we will be able to manage all the bot from studio as a NO-CODE-BOT

Thanks for your help

Regards

Mat

francoisno:
cc @msansen-esncf and @broxmik (still to do: media checkbox/radio to enable-disable the mechanism).


vsct-jburet:
The idea is nice but the behaviour is not backward compatible. So you need to add a checkbox first. ;) But then, is it not better to add a carousel component in tock ui ?

PR #1249, opened 2021-04-05
resolves #1246 Web: connector bridge

### LINKED ISSUE #1246: [Connectors] Multi-configuration Web connector
For bots with multiple _configurations_ for different frontends (even different pages on the same website), it would be nice, not to have to instantiate multiple Web connectors - multiple endpoints - by passing the configurationId to the connector.

In Tock Studio, the connector would be instantiated inside a configuration - as usual. By default the configuration would remain this "parent" configuration.

When calling the connector with an explicit configurationId (an additional parameter, as header?), the message would be routed to that configuration, provided it exists in the connector parent application.

In Tock Studio, enabling the _multi-configuration_ feature would be an additional checkbox in the connector setup.

_By the way this feature could generalize to any connector - yet it is most useful for the Web connector._

francoisno:
Very simple implementation : no UI modifications, add `connectorId` to Web request to transfer to another connector.

No security checks (namespace etc.), but feature is disabled by default : `booleanProperty("tock_web_connector_bridge_enabled", false)`

### DIFF
--- a/bot/connector-web-model/pom.xml
+++ b/bot/connector-web-model/pom.xml
@@ -24,6 +24,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>tock-bot-connector-web-model</artifactId>
+    <name>Tock Bot Web Connector Model</name>
     <dependencies>
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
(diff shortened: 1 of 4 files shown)


vsct-jburet (approved):
ok for me

PR #1293, opened 2021-09-09
[Feature issue-1263] : Implement story ending rule for built-in stories

amineSt:
Signed-off-by: Amine <amine.cherfi@smart-tribune.com>


samail:
@francoisno @vsct-jburet 
Hi, here is the expected PR from an account within our organization. 

We were not able to unit test BotApiHandler. Since this class was not tested yet, we tried to implement it but it seems that we are not able to Mock Websocket tunnel because of missing dependency injection of "WebSocketController". We remain available to discuss this if needed. 

Thank you

vsct-jburet:
Thank you for the PR. I have slighty refactored the code to add a test :) Merged

PR #1340, opened 2022-03-15
update markdown

pi-2r:
update of the href tag with `target="_blank" rel="noopener noreferrer"`.

### DIFF
--- a/bot/connector-web/src/main/kotlin/WebMarkdown.kt
+++ b/bot/connector-web/src/main/kotlin/WebMarkdown.kt
@@ -70,6 +70,12 @@ object WebMarkdown {
             if (render.contains("~~")) {
                 render = extractAllDataWithRegex(regex, render)
             }
+            if (render.contains("<a href=")) {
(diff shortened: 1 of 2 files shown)


francoisno (approved):

Fabilin:
While I understand it's getting standard to always open links in new tabs, I have a feeling some TOCK users could want the default HTML behaviour... Maybe this could be configurable somewhere?

francoisno:
> While I understand it's getting standard to always open links in new tabs, I have a feeling some TOCK users could want the default HTML behaviour... Maybe this could be configurable somewhere?

I think you're right, it would be nice to be able to remain in the same tab as an option. Yet opening new tabs from a chatbot seems definitely the de facto standart/requirement. I suggest we provide the "same target" option later on (as a new issue).

Pyrofab (approved):

Fabilin (approved):

vsct-jburet (approved):

Issue #1168, opened 2021-01-26
[Studio] Basic Q&A management proposal


jygdev:
Hi,

As Arkea's conversational team we receive many requests for basic Q&A chatbots creation.
It is currently hard for people not familiar with Tock or conversational concepts to create this kind of bot by themselves into the Studio.
We would like to propose a new user experience to be included into the studio to handle such use cases and give our customers more autonomy with their basic Q&A bots creation and content management.

We designed few mockups (see attachment) of how we could add this to the studio based on needs we have collected from our customers.
We would be happy to discuss further this proposal with anyone interested before strating developing this new feature.
Feel free to ask questions if everything isn't clear enough in the mockups.

Regards,

[TOCK FAQ.pdf](https://github.com/theopenconversationkit/tock/files/5874205/TOCK.FAQ.pdf)

vsct-jburet:
merged, thank you !

PR #1360, opened 2022-04-22
1168 faq UI : with new roles support

djowalker:
To see clearly the differences : you should first merge #1168 with #1359, then with the diff you could see the main differences with new roles :

In TockUserRole :
`faqNlpUser` 
`faqBotUser`

Some of the endpoints restrictions have changed , there are many shared endpoints for botUser and faqBotUser and also nlpUser and faqNlpUser.
The menu display are now working according to the role the user have.

faqNlpUser have acces to :
- Test
- FAQ Training

faqBotUser have access to :
- Test
- Analytics
- FAQ Training
- FAQ Management


bot/admin/server/src/main/kotlin/BotAdminService.kt
```diff
@@ -1086,10 +1063,18 @@ object BotAdminService {
         )
     }
 
+    /**
+     * Checks and save the story
+     * @param namespace
+     * @param : story : botStoryDefinitionConfiguration
+     * @param: user : userLogin
+     * @param: createIntent : intent can be created out of the method
```
djowalker, line 1071: createdIntent

bot/admin/web/.gitconfig
```diff
@@ -0,0 +1,3 @@
+[url "https://"]
+	insteadOf = git://
```
djowalker, line 2: Ne devrait pas faire partie de la PR

PR #1370, opened 2022-05-17
Teams change action types

pi-2r:
Following analysis with a Microsoft expert, we need to test the "ImBack" solution (https://docs.microsoft.com/en-us/microsoftteams/platform/task-modules-and-cards/cards/cards-actions?tabs=json#action-type-imback) by adding the value element.

I only need a snapshot with these changes to test.


vsct-jburet (approved):

pi-2r:
@vsct-jburet or @francoisno the patch fixes the display bug on TEAMS.
This patch can be pushed on the master branch.

PR #1376, opened 2022-06-14
[FAQ] Enhancement and updates

djowalker:
- Full frontend refactoring about FAQ
- Introduce FAQSettings to use redirection (for example satisfaction on all FAQ)
- More unit tests
- Fixing search trouble when a classified sentences is deleted, refactoring and testing the code for aggragation search and add a filter ClassifiedSentences status because the BuildWorker update the sentences each 1s.
- Adding a fix to work on DocumentDB when using aggregation pipeline request for FAQ
- Show MISSING_UTTERANCE in frontend when a FAQ has no more classified sentences associated.
- FAQ will use story name and description instead of the actual intention name and description : to avoid trouble with shared intents
- Some more unit test in frontend
- Add Display Unknown sentences only in FAQ Training


vsct-jburet:
@djowalker is it still a draft?

djowalker:
3 more things that are going to be added in this PR :
- FAQ will use story label instead of the actual intention name
- Some more unit test in frontend
- Add Display Unknown sentences only in FAQ Training

djowalker:
@vsct-jburet ready for review

djowalker:
@vsct-jburet : There was a trouble with the tests in Messenger, but you upgrade the version recently so i added the fix from Master in this branch

vsct-jburet (approved):

vsct-jburet:
LGTM, thank you !

PR #1425, opened 2022-12-08
Resolves #1409 Studio: Sort stories by last updated date

### LINKED ISSUE #1409: Studio : filter intents and stories by lastupdated date
We would like to have the possibility to filter stories and intents by their lastupdated date. 

To do so we were thinking about adding a filter button "modified after or modified before" next to the "Category" button.

 This new functionality would allow us to check more easily which stories or intent might need an update.

Matomatt:
In stories & Answers / Stories (url build/story-search) :
- Add a new column with the last updated date for the stories
- Add a toggle to sort stories by last updated date
- The search bar now searches in story name, id and intent


vsct-jburet:
LGTM. Thank you !

Issue #1457, opened 2023-02-22
Rocket.Chat connector gets an exception during connection


serjk:
I've tested Rocket.Chat connector. During connection process gets an exception:
`Exception in thread "DefaultDispatcher-worker-1 @coroutine#8" java.lang.NoSuchMethodError: 'boolean kotlinx.coroutines.channels.Channel.isFull()'
	at chat.rocket.core.internal.realtime.socket.Socket$onMessage$$inlined$let$lambda$2.invokeSuspend(Socket.kt:349)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:664)
	Suppressed: kotlinx.coroutines.DiagnosticCoroutineContextException: [CoroutineId(8), "coroutine#8":StandaloneCoroutine{Cancelling}@72c89528, Dispatchers.Default]`

The problem is that rocket.chat connector uses rocket.chat.sdk which requires the old coroutine lib(v.1.1.1).
https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/blob/app-3.5.1/dependencies.gradle

Tock uses newer coroutine lib version (v.1.6.4)
https://github.com/theopenconversationkit/tock/blob/a56f8606346181528f676b6dc503194cbf89fb61/pom.xml#L35

Rocket.Chat.sdk contains the fix but only on develop branch.There is no release to publish to mvn repository.  
https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/blob/develop/dependencies.gradle#L5

Does someone have any ideas how to fix it?

PR #1458, opened 2023-02-23
Downgrade coroutines lib version in rocket.chat connector

serjk:
The latest rocket.chat.sdk requires the old coroutines lib version.
#1457

### DIFF
--- a/bot/connector-rocketchat/pom.xml
+++ b/bot/connector-rocketchat/pom.xml
@@ -30,6 +30,7 @@
 
     <properties>
         <rocket>3.5.1</rocket>
+        <rocket-coroutines>1.1.1</rocket-coroutines>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


vsct-jburet:
Thank you for the PR. I do not like the idea to have to coroutine versions in the classpath :(
Do you kown if there is a more recent version of rocket.chat kotlin sdk ? Or may be it's easier do just add a simple http client for rocket.chat api ?

serjk:
@vsct-jburet I agree that is not the best solution....
Rocket.Chat.sdk contains the fix but only on develop branch. The latest published lib doesn't contain required changes.
Develop branch: https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/blob/develop/dependencies.gradle#L5
Latest release: https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/blob/982d97a169f5e7b613ed0bd3c2554d3fb55fff07/dependencies.gradle#L5

I've also written to a guy who published the last release version to maven central. He doesn't work under this project and said that this project has been deprecated.

PR #1462, opened 2023-02-28
[ANALYTICS - FLOW] Save a snapshot before switching ending story

assouktim:


bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -247,63 +244,62 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
             logger.debug { "dialog saved $userTimeline" }
         }
 
-        executor.executeBlocking {
```
vsct-jburet, line 250: This is not a good idea to remove this executeBlocking block. Why doing this ?

assouktim: Due to data state synchronization issue in coroutines.

Now it's fixed. I implemented the following solution: https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#thread-confinement-coarse-grained
I

vsct-jburet: I still don't like the PR.. I suggest we start a call about this :)

assouktim: Yes, I'm available for the rest of the week. We can talk about this RP.
What about you?

vsct-jburet: I'v sent a message on gitter ;)

assouktim: Following our discussion, the code has been updated

vsct-jburet (approved):

PR #1463, opened 2023-03-01
fixes #1459 connector_web: Fix SSE compatibility, update property name and documentation

### LINKED ISSUE #1459: [Bug] [Web Connector] Markdown implementation fails in multiple scenarios
While attempting to use the `allow_markdown` feature introduced in v21.9.6, we noticed that it did not mesh very well with other tock features - most notably:
1. Only the simplest text messages get formatted, simply adding a button to a message prevents markdown from working
2. SSE completely cancels out markdown formatting

The `allow_markdown` system/environment property is also the only one to not be prefixed with `tock_`, making it stand out somewhat.

Fabilin:
Fixes #1459 

This PR:
- moves the web message processing logic into its own class, used by both the default web connector callback and the SSE channels handler
- adds processing logic for already constructed `WebMessage`s
- renames the property from `allow_markdown` to `tock_web_enable_markdown` to be more in line with other TOCK properties
    - the old property is kept as a fallback for backward compatibility
- updates the README to mention the markdown feature and its limitations

### DIFF
--- a/bot/connector-web/README.md
+++ b/bot/connector-web/README.md
@@ -129,7 +129,12 @@ response:
 
 A simple [Swagger descriptor](./Swagger_TOCKWebConnector.yaml) of the rest service is provided¬.
 
-## Server-sent events (SSE)
+## Additional features
+
+Several features can be optionally used with the Web Connector. Some require specific properties to be set, either
+as a Java system property or as an environment variable (system property takes precedence).
+
+### Server-sent events (SSE)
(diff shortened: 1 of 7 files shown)


kbvi (approved):

pi-2r (approved):

vsct-jburet (approved):

vsct-jburet:
Thank you !

PR #1469, opened 2023-03-31
fixes #1433 bug with actions when updating Media card

### LINKED ISSUE #1433: [Bug] [Tock Studio] Bugs with actions when updating the Tock version
Since we updated the Tock version, the actions buttons in the Media Card are not corrects. We have the same problem on the preprod and prod environment  : we have the correct number of buttons but the same text and same link on every button. It is always the text + link of the last button of the media card.
The buttons are correct when we use the bot, they are only incorrect in the backoffices

image

image

correi-f:
fixes #1433 bug with actions when updating Media card

It seems that Object.create and Object.assign do not create an internalId as expected. All media actions have the same name, that's why we see always the same input duplicated when multiple actions.


AICHOUN-C (approved):
It's fine

Issue #1471, opened 2023-04-07
[Verticles] Reorganize endpoints per business


djowalker:
A lot of endpoint are present in AdminVerticle and BotAdminVerticle.
It could be interesting for developers to have it more stuctured per businesses.

For exemple : 
per FAQ
per I18n
per analytics.

etc.

djowalker:
The proposal was following breaking changes evolves with generic error handling per Verticles.
However the idea to cut the verticles into business is still really interesting.

But there is still an idea of Parent and child Verticle for Bot Admin Server.

PR #1474, opened 2023-04-12
New way of verticle errors handling and faq shared intents

### LINKED ISSUE #1470: [Verticles] Generalize errors and avoid 500 errors without details
For developers it would be interesting to help them to create only RestException and structure more properly the code with HTTP Responses Exception when creating new endpoint and verticles instead of the default error(), notfound() methods.

The idea is hava a manager who takes a specific exception and send a RestException.
This allow to have a better error system with HTTP code and also with business error codes for instance.

### LINKED ISSUE #1472: [FAQ] delete shared intents
To delete shared intent, yet there is an NotImplementedErrorException().

djowalker:
fixes #1470 
fixes #1472 for faq shared intent deletion, but also for faq creation with shared intent or unwanted shared intent.


djowalker:
I have some more elements to manage about the authentication providers and the CAS sample example.

djowalker:
Everything seems ok however that is big change.
Then a will follow this one and is about #1471

djowalker:
New pull request dedicated only for #1470 is created

PR #1480, opened 2023-04-25
fix #1479 [export/import stories] improve export service - get label up to date

### LINKED ISSUE #1479: [bug] [export stories] stories have labels that are not up-to-date - export service

IdrissHam:


vsct-jburet:
@IdrissHam What exactly is the purpose of this PR ?

IdrissHam:
> @IdrissHam What exactly is the purpose of this PR ?

When exporting stories, we detected an inconsistency between the data displayed on the user interface (UI) and the exported data. After a thorough analysis, it became evident that the labels in the collection used as the basis for export were not being updated.

The solution implemented involves applying the same mechanism used for the story search service to the export service.

vsct-jburet:
@IdrissHam this is WAD. Today you need to import the labels before the stories - if you prefer to implement a full import (ie stories + Label) you need to add a new feature and not to modify the current behaviour.

PR #1481, opened 2023-04-25
1478 [story search] improvement of the search bar with several words

pi-2r:


bot/storage-mongo/src/main/kotlin/StoryDefinitionConfigurationMongoDAO.kt
```diff
@@ -195,6 +195,25 @@ internal object StoryDefinitionConfigurationMongoDAO : StoryDefinitionConfigurat
         return col.find(and(Namespace eq namespace, BotId eq botId)).toList()
     }
 
+    fun customRegexToFindWord(textSearch: String):String {
```
correi-f, line 198: We can add a limit to avoid performance issue, like 10 words max maybe
```suggestion
    fun customRegexToFindWord(textSearch: String) = if (textSearch.trim().isEmpty()) {
        ""
    } else {
        textSearch
            .trim()
            .split("\\s+".toRegex())
            .filter { it.isNotEmpty() }
            .joinToString("", "^", "\$", 10) { wordTextSearch ->
                "(.*?(${allowDiacriticsInRegexp(wordTextSearch)})[^\$]*)"
            }
    }
```

PR #1498, opened 2023-06-08
[Studio] fix #1497 Allow access to namespace setting tab to more users / roles

### LINKED ISSUE #1497: [Tock Studio]  Switching between namespace for any user (Setting hidden)
Hi,
Namespace are accessible only from Settings > Namespaces (tab), which is hidden for non admin user. Non admin users can have rights one different applications located in different namespaces, actually they can't switch namespaces.

**Use case**
As a tock user (any role), allowed on different namespaces I need to be able to switch from one namespace to an other one. So that I can use the studio on all the applications on all namespaces that I have access to.

@rkuffer from Arkea team has developed the fix, I will push the PR.

Benvii:
Fixes #1497, make the Settings > Namespace tab accessible for all tock users as they may have access to several namespace without behind admin.

More details in the issue.


bot/admin/web/src/app/bot-admin-app.component.ts
```diff
@@ -89,7 +89,7 @@ export class BotAdminAppComponent implements AuthListener, OnInit, OnDestroy {
         title: 'Settings',
         icon: 'settings-outline',
         link: '/configuration',
-        hidden: !this.state.hasRole(UserRole.admin)
+        hidden: !this.state.hasRole(UserRole.botUser) && !this.state.hasRole(UserRole.faqBotUser)
```
Benvii, line 92: We should open this menu entry to any user without restrictions. All users need to be able to switch namespace.

I would simply set `hidden: false` with a comment indicating that all users can switch namespace.

bot/admin/web/src/app/core/bot-core.module.ts
```diff
@@ -32,8 +32,8 @@ export class BotCoreConfig implements CoreConfig {
   roleMap: Map<UserRole, string[]> = new Map([
     [UserRole.nlpUser, ['/nlp']],
     [UserRole.faqNlpUser, ['/faq/training']],
-    [UserRole.faqBotUser, ['/faq/manangement', '/build/i18n']],
-    [UserRole.botUser, ['/build']],
+    [UserRole.faqBotUser, ['/faq/manangement', '/build/i18n', '/configuration']],
+    [UserRole.botUser, ['/build', '/configuration']],
```
Benvii, line 36: Good 👍️ but nlpUser and faqNlpUser need also access to the configuration route so that thet can train all their applications in all namespaces

Benvii (commented):
I think we can open more the setting menu, it should be accessible for all users.

Let's say I'm an `nlpUser` user, I have access to BotA in namespaceA and BotB in namespaceB, I want to be able to qualify sentences / train BotA and BotB so I need access to the configuration/settings menu and namespace tab also.

bot/admin/web/src/app/bot-admin-app.component.ts
rkuffer: Done

bot/admin/web/src/app/core/bot-core.module.ts
rkuffer: Done

vsct-jburet (approved):

PR #1578, opened 2024-02-07
resolve #1577 Admin : Add creationDate into export sentences dump

### LINKED ISSUE #1577: [Tock Studio] Creation date missing in sentences data export
In order to analyze data of BOT IV in production, we need to have the creationDate in sentences data export
Is it possible to add the creationDate ?

asus007:


nlp/front/shared/src/main/kotlin/codec/SentencesDump.kt
```diff
@@ -32,6 +33,7 @@ data class SentencesDump(
 data class SentenceDump(
     val text: String,
     val intent: String,
+    val creationDate : Instant,
```
vsct-jburet, line 36: please add `val creationDate : Instant = now().truncatedTo(ChronoUnit.MILLIS)` because deserialization of old dumps will fail

Fabilin: Is this actually ever deserialized directly from the database? I was under the impression that this class was only instantiatied by the `ApplicationCodecService`.

vsct-jburet: This is deserialized from the uploaded json file I think :)

vsct-jburet: @Fabilin  any news ?

PR #1643, opened 2024-06-21
[DERCBOT-1071] Prevent duplication of namespaces

assouktim:
After authentication on Tock, when registering the user, if the namespace we are trying to create already exists (comparison now case insensitive), I take what is already in the database, so that the name is not duplicated.

At the time of registration via Studio/Settings/Namespace, the namespace existence check was modified to ignore case when comparing.

### DIFF
--- a/nlp/front/storage-mongo/src/main/kotlin/UserNamespaceMongoDAO.kt
+++ b/nlp/front/storage-mongo/src/main/kotlin/UserNamespaceMongoDAO.kt
@@ -23,13 +23,16 @@ import ai.tock.nlp.front.shared.user.UserNamespace_.Companion.Login
 import ai.tock.nlp.front.shared.user.UserNamespace_.Companion.Namespace
 import ai.tock.nlp.front.shared.user.UserNamespace_.Companion.Owner
 import com.mongodb.client.MongoCollection
+import com.mongodb.client.model.Filters
 import com.mongodb.client.model.ReplaceOptions
+import org.bson.conversions.Bson
 import org.litote.kmongo.and
 import org.litote.kmongo.ensureIndex
 import org.litote.kmongo.ensureUniqueIndex
 import org.litote.kmongo.eq
 import org.litote.kmongo.getCollection
 import org.litote.kmongo.setValue
+import java.util.regex.Pattern
 
 object UserNamespaceMongoDAO : UserNamespaceDAO {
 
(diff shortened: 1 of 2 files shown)


nlp/front/storage-mongo/src/main/kotlin/UserNamespaceMongoDAO.kt
```diff
@@ -69,5 +70,5 @@ object UserNamespaceMongoDAO : UserNamespaceDAO {
         col.countDocuments(and(Login eq user, Namespace eq namespace, Owner eq true)) == 1L
 
     override fun isExistingNamespace(namespace: String): Boolean =
-        col.countDocuments(Namespace eq namespace) != 0L
+        col.countDocuments(Filters.regex(Namespace.name, "^$namespace$", "i")) != 0L
```
vsct-jburet, line 73: I do not understand with the namespace is not case sensitive. Also this code allows injection :). 

It would be better to optionally format the namespace to lower case when the namespace is created. What is your use case ?

assouktim: yes, it's true, I can prevent this risk by using Pattern.quote : 
`val escapedNamespace = Pattern.quote(namespace)
col.countDocuments(Filters.regex(Namespace.name, "^$escapedNamespace$", "i"))`

The use case is described here : https://github.com/theopenconversationkit/tock/issues/1644

vsct-jburet: LGTM. Yes at least quote the namespace. And add a option to disable the feature (for backward compatibility)

assouktim: If I've understood correctly, you want me to add an environment var to enable or disable this fix.
This comparison change has been made to correct an anomaly, not doing this fix means that we accept to have 'app' and 'App' as two different namespaces. So adding an option to enable or disable this comparison doesn't seem justifiable to me.

What do you think?

vsct-jburet: Ok I understand better the use case, but :

- why changing isExistingNamespace that is not called by AdminTockUserListener ?

- you also have to change  namespaceDAO.getUsers(namespace) call (AdminTockUserListener#46) to handle case sensitive issue

assouktim: - The isExistingNamespace function is called when a namespace is created, so we check that the namespace exists before creating it.
- Yes, that's a good point. I'm working on it

vsct-jburet (approved):

PR #1649, opened 2024-06-24
[DERCBOT-1064] Tag the Gen AI Rag answer to filter dialogs

assouktim:


bot/admin/server/src/main/kotlin/BotAdminService.kt
```diff
@@ -244,6 +242,18 @@ object BotAdminService {
                         }
                     )
                 }
+
+                if(query.isGenAiRagDialog == true) {
```
Benvii, line 246: Hum filter should be done at mongo request (so at DAO level), not after because it will impact the paginated notion of that query.

`DialogsSearchQuery` is a `PaginatedQuery` which means that the user can request a specific amount of results. Let's say we request 5 dialogs as it will be limited without the filter before [here](https://github.com/CreditMutuelArkea/tock/blob/master/bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt#L510C22-L510C27) we will request 5 dialog then filter the genAI ones ... So we won't get the latest 5 genAI dialogs.

Can you move this filtering into the [DAO here](https://github.com/CreditMutuelArkea/tock/blob/master/bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt#L608) ?

Benvii (changes requested):

bot/admin/server/src/main/kotlin/BotAdminService.kt
assouktim: yes I agree, I wasn't paying attention. thanks for the comment

Benvii (approved):
Ok for me for this new filter.

Benvii (approved):

vsct-jburet (approved):

PR #1652, opened 2024-06-25
[DERCBOT-1060] Delete obsolete FAQ roles

assouktim:


Benvii:
Hi,
Can you create an issue to describe the change and why we are making it ?

I have added this subject for the next "Tock Community Virtual Coffe ☕️🤖" on Thursday, 4 July⋅4:00 – 5:00pm. This is clearly a huge breaking change (if someone is using those role in the community), I think that we should first deprecate the roles with warning then remove them.

I'm not sure it's the best approach be we could wrap the auth provider "[toTockUser()](https://github.com/theopenconversationkit/tock/blob/master/shared/src/main/kotlin/security/auth/TockAuthProvider.kt#L61)" and remap the following role (and remove duplicates) : 

- faqBotUser => botUser
- faqNlpUser => nlpUser

It's call here and could be temporary wrapped here with a warning indicating : 
https://github.com/theopenconversationkit/tock/blob/master/shared/src/main/kotlin/vertx/WebVerticle.kt#L899

assouktim:
issue : https://github.com/theopenconversationkit/tock/issues/1656

Benvii (approved):

AurelienSylvan (approved):

Benvii (approved):
Ok but it will still break for Authentication providers that return those roles as as they will be refused by the Bot Admin API.

I was mentioning wrapping them mapping it to BotUser and NlpUser after the authentication provider is called in my comment : https://github.com/theopenconversationkit/tock/pull/1652#issuecomment-2193943544

Benvii:
We should also create a ticket to remove this part https://github.com/theopenconversationkit/tock/pull/1652/files#diff-bfcbf9099787d7fe4cf1aae4918633ca6ac816ece6f83d1fd8d1a0ccbcd857e8R901  in a future version.

PR #1657, opened 2024-06-27
[Tock Studio] Filter dialogs with rag responses + Create new Faq from rag response

rkuffer:
This PR adds :
- An option on analytics/search to list only dialogs containing Rag responses
- An option on analytics/search to create a new Faq entry based on a question/answer pair from a Rag response

Depends on https://github.com/theopenconversationkit/tock/pull/1649

tock-faq-from-rag


assouktim (approved):
Changes already approved here : 
https://github.com/CreditMutuelArkea/tock/pull/370

Benvii (approved):

PR #1683, opened 2024-07-11
#1682 whatsAppCloudTextParameterTemplate - fix problem with spanish

asus007:


bot/connector-whatsapp-cloud/src/main/kotlin/WhatsAppCloudBuilder.kt
```diff
@@ -328,7 +328,7 @@ fun <T : Bus<T>> T.whatsAppCloudTextParameterTemplate(
     typeParameter: CharSequence?,
     textButton: CharSequence?
 ): TextParameter = TextParameter(
-    type = ParameterType.valueOf(translate(typeParameter).toString()),
+    type = ParameterType.valueOf((typeParameter).toString()),
```
Fabilin, line 331: Why not just take the `ParameterType` as argument to the method instead of going through the valueOf?

Fabilin (approved):

vsct-jburet (approved):

PR #1700, opened 2024-07-23
[Tock studio] Bot and Nlp modules merge

rkuffer:
- Nlp module directories are moved into their Bot module equivalents (previously reflected by symbolic links).  
- “nlp/admin/web” is completely removed.
- Nlp components and modules no longer required are removed


Benvii (approved):

assouktim (approved):
Thanks !

PR #1727, opened 2024-08-28
[Tock Studio] Update app header after current application label change

rkuffer:
When changing the current application label, the app header application menu doesn't update. This fixes the bug


assouktim (approved):
Thanks

morgandiverrez (approved):

AurelienSylvan (approved):

Benvii (approved):

PR #1731, opened 2024-08-29
[Indexing tools] Improvements (Use of Secret Key Structure, Csv field size)

assouktim:


gen-ai/orchestrator-server/src/main/python/tock-llm-indexing-tools/embeddings_azure_openai_settings.example.json
Benvii, line 1: - [ ] We should clearly refer to the orchestrator swagger or the python model source files in the REAME.md to document this configuration file format (so that people not using Azure OpenAI can also use it). We currently only have this `it uses the embeddings constructor from the orchestrator module, so JSON file shall follow corresponding format`, without any link or pointer to the model in question.
- [ ] We also need to document in the README.md the target database configuration so that people can index their data not just to their local Open Search instance

Benvii (changes requested):
Thank a lot for this PR, I think this PR makes the tools more usable it still needs a few small update and it will be good for merge. Thanks in advance.

gen-ai/orchestrator-server/src/main/python/tock-llm-indexing-tools/embeddings.json.example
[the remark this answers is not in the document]
assouktim: Yes, it's true that this is a possible risk, but it's difficult to target this json file, as the file name can be called anything, and can be in any directory. 
To limit this risk, I've ignored this file with a regexp to cover the most obvious cases

gen-ai/orchestrator-server/src/main/python/tock-llm-indexing-tools/embeddings_azure_openai_settings.example.json
assouktim: all done

morgandiverrez (approved):

Benvii (approved):
Thanks 🙂

PR #1738, opened 2024-09-13
[ADMIN] resolves #1737 Add graduation to application features

### LINKED ISSUE #1737: [Application Feature] Add gradual activation depending on user identifier
Add the possibility to add a graduation on application features in Tock studio to activate the feature only for a percentage of users.
The graduation can be added in the feature state saved in Mongo database as an optionnal Int with a value between 0 and 100
Expected behaviour :
- FT DISABLED : DISABLED regardless of graduation saved
- FT ENABLED and no graduation (null value) : ENABLED
- FT ENABLED and graduation = 0 => DISABLED
- FT ENABLED and graduation = 100 => ENABLED
- FT ENABLED and graduation = x (between 1 and 99) => if(hash(userId) < x) ENABLED else DISABLED
Where hash is a function which result is a number between 0 and 100 from the userId String

gsoumier:


bot/storage-mongo/src/main/kotlin/FeatureMongoDAO.kt
```diff
@@ -72,17 +76,33 @@ internal class FeatureMongoDAO(private val cache: FeatureCache, private val col:
...
 
-    private fun isEnabled(feature: Feature): Boolean {
-        val now = ZonedDateTime.now(internalDefaultZoneId)
+    private fun isEnabled(feature: Feature, userId: String?): Boolean {
+        return feature.enabled
+                && isEnableForDate(feature, ZonedDateTime.now(internalDefaultZoneId))
+                && isEnableForUser(feature, userId)
+    }
+
+    private fun isEnableForUser(feature: Feature, userId: String?): Boolean {
```
vsct-jburet, line 89: Here you have to set a a default value for userId, to ensure backward compatibility : `userId: String? = null`

vsct-jburet (approved):

PR #1741, opened 2024-09-20
[indexing-tools]script pre processing export faq tock studio json yo ready-to-index csv

morgandiverrez:


assouktim:
we no longer need to do it this way, because on the Tock studio, thanks to this [PR](https://github.com/theopenconversationkit/tock/pull/1765), we can now export FAQs in csv format, and they're ready to be ingested.

PR #1762, opened 2024-10-10
[DERCBOT-1159] Immediate deletion of secrets in the secret manager

assouktim:
This pull request introduces several changes to enhance the security and observability features of the bot administration services. The most significant updates include the addition of secret deletion functionality and the integration of observability settings.

### Security Enhancements:

* Added `deleteSecret` method in `SecurityUtils` to handle secret deletion, with support for different secret types and providers (`gen-ai/orchestrator-core/src/main/kotlin/ai/tock/genai/orchestratorcore/utils/SecurityUtils.kt`).
* Implemented the `deleteSecret` method in `SecretManagerService` interface and its various implementations, including AWS, GCP, and environment-based secret managers (`shared/src/main/kotlin/security/SecretManagerService.kt`, `util/aws-tools/src/main/kotlin/ai/tock/aws/secretmanager/AwsSecretManagerService.kt`, `util/gcp-tools/src/main/kotlin/ai/tock/gcp/secretmanager/GcpSecretManagerService.kt`, `util/env-tools/src/main/kotlin/ai/tock/env/secretmanager/EnvSecretManagerService.kt`) [[1]](diffhunk://#diff-8a58a26a135632242cf9bb6da9cb5ca9db35527c5d98e259f2d39565eaea406fR101-R105) [[2]](diffhunk://#diff-ff91519f484b8d12a004c580c63b59184dc4b8b2e88a756481d48563ce21ae49R196-R208) [[3]](diffhunk://#diff-a924714744b82c48bb1b1c339af0d084b38754b5355ef34d817bc0acd48da36eR137-R146).

### Observability Integration:

* Added imports for `LangfuseObservabilitySetting` and `SecurityUtils` in `BotAdminService`, `ObservabilityService`, and `RAGService` to support observability settings and secret management (`bot/admin/server/src/main/kotlin/BotAdminService.kt`, `bot/admin/server/src/main/kotlin/service/ObservabilityService.kt`, `bot/admin/server/src/main/kotlin/service/RAGService.kt`) [[1]](diffhunk://#diff-247695b29ea9fe29753c2a141fd2c331950837b442b867c2b0927360e27fdedeR41-R42) [[2]](diffhunk://#diff-fbc899c246301bb38c0b2532f774260483a788be7bc481677d9891cb703bcc94R23-R24) [[3]](diffhunk://#diff-1a32c8cf0c275ddb0ad9e9515b485b4452395d5c0d156db82ae2c4c8c720486fR26).

### Configuration Deletion Updates:

* Enhanced configuration deletion methods across various services (`BotAdminService`, `ObservabilityService`, `RAGService`, `SentenceGenerationService`, `VectorStoreService`) to ensure associated secrets are also deleted (`bot/admin/server/src/main/kotlin/BotAdminService.kt`, `bot/admin/server/src/main/kotlin/service/ObservabilityService.kt`, `bot/admin/server/src/main/kotlin/service/RAGService.kt`, `bot/admin/server/src/main/kotlin/service/SentenceGenerationService.kt`, `bot/admin/server/src/main/kotlin/service/VectorStoreService.kt`) [[1]](diffhunk://#diff-247695b29ea9fe29753c2a141fd2c331950837b442b867c2b0927360e27fdedeL1144-R1171) [[2]](diffhunk://#diff-fbc899c246301bb38c0b2532f774260483a788be7bc481677d9891cb703bcc94R67-R75) [[3]](diffhunk://#diff-1a32c8cf0c275ddb0ad9e9515b485b4452395d5c0d156db82ae2c4c8c720486fR63-R70) [[4]](diffhunk://#diff-4d669fdbbff164a0c0a2b9b43c1b1ed1790431d905424cf25d71bea55a90ce26L55-R61) [[5]](diffhunk://#diff-c5f3084437f8670e52ae291382dadf5f76e0abb1902bbcaa921257d2c175a2bdR67-R72).

These changes collectively enhance the security and observability capabilities of the bot administration services by ensuring that secrets are properly managed and deleted when configurations are removed.

### DIFF
--- a/bot/admin/server/src/main/kotlin/BotAdminService.kt
+++ b/bot/admin/server/src/main/kotlin/BotAdminService.kt
@@ -38,6 +38,8 @@ import ai.tock.bot.admin.dialog.*
 import ai.tock.bot.admin.kotlin.compiler.KotlinFile
 import ai.tock.bot.admin.kotlin.compiler.client.KotlinCompilerClient
 import ai.tock.bot.admin.model.*
+import ai.tock.bot.admin.service.ObservabilityService
+import ai.tock.bot.admin.service.RAGService
 import ai.tock.bot.admin.story.*
 import ai.tock.bot.admin.story.dump.*
 import ai.tock.bot.admin.user.UserReportDAO
@@ -49,6 +51,8 @@ import ai.tock.bot.engine.dialog.DialogFlowDAO
 import ai.tock.bot.engine.feature.FeatureDAO
 import ai.tock.bot.engine.feature.FeatureState
 import ai.tock.bot.engine.user.PlayerType
+import ai.tock.genai.orchestratorcore.models.observability.LangfuseObservabilitySetting
+import ai.tock.genai.orchestratorcore.utils.SecurityUtils
 import ai.tock.nlp.admin.AdminService
 import ai.tock.nlp.core.Intent
 import ai.tock.nlp.front.client.FrontClient
@@ -58,6 +62,8 @@ import ai.tock.nlp.front.shared.config.ClassifiedSentenceStatus.model
 import ai.tock.nlp.front.shared.config.ClassifiedSentenceStatus.validated
 import ai.tock.shared.*
 import ai.tock.shared.security.UserLogin
+import ai.tock.shared.security.key.HasSecretKey
+import ai.tock.shared.security.key.SecretKey
 import ai.tock.shared.vertx.WebVerticle.Companion.badRequest
 import ai.tock.translator.*
 import com.github.salomonbrys.kodein.instance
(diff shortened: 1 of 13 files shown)


bot/admin/server/src/main/kotlin/BotAdminService.kt
```diff
@@ -1141,23 +1145,30 @@ object BotAdminService {
...
+                SecurityUtils.deleteSecret((config.setting as LangfuseObservabilitySetting).secretKey)
+            }
         }
 
         // delete the Vector Store configuration
-        vectorStoreConfigurationDAO.findByNamespaceAndBotId(app.namespace, app.name)?.let {
-            vectorStoreConfigurationDAO.delete(it._id)
+        vectorStoreConfigurationDAO.findByNamespaceAndBotId(app.namespace, app.name)?.let { config ->
+            vectorStoreConfigurationDAO.delete(config._id)
+            SecurityUtils.deleteSecret(config.setting.password)
```
AurelienSylvan, line 1171: At which layer are the Mongo transaction handled ? If the secret deletion did not happen, will the mongo deletion be rollback ? What happens if we replay this deletion process because ot partially failed ? Is it idempotent ?

gen-ai/orchestrator-core/src/main/kotlin/ai/tock/genai/orchestratorcore/utils/SecurityUtils.kt
```diff
@@ -83,5 +83,31 @@ object SecurityUtils {
...
+            if(secretMangerService.isSecretTypeSupported(secret)) {
+                return secretMangerService.deleteSecret((secret as NamedSecretKey).secretName)
+            }else{
+                throw IllegalArgumentException("The secret manager provider type '${secret.type}' is not supported by " +
+                        "the instantiated service ${secretMangerService::class.simpleName}.")
+            }
+        } catch (e: Exception) {
+            logger.warn("The secret has not been deleted.", e)
+            // Do not block treatment if it fails.
+        }
```
AurelienSylvan, line 110: Very good point for the non-blocking failure. So why theses immediately caught exceptions ?
I think the exception handling should be made in higher layers, letting the consumer decide what to do in case of failure, depending on the context. WDYT ?

assouktim: Yes, you're absolutely right, but today on tock we're using this service in 10 places, and they're all iso in terms of behavior. In fact, that's exactly what we did for `fetchSecretKeyValue`.

bot/admin/server/src/main/kotlin/BotAdminService.kt
assouktim: I knew the operation could succeed on the database, but not on the cloud, but I didn't think it was a big deal, because you'd end up with an orphaned secret. What do you think @Benvii? Is it necessary to go back (redo an insertion) on the MongoDB deletion if the secret deletion didn't go well?

bot/admin/server/src/main/kotlin/service/ObservabilityService.kt
```diff
@@ -62,8 +64,15 @@ object ObservabilityService {
...
+
         logger.info { "Deleting the Observability Configuration [namespace: $namespace, botId: $botId]" }
-        return observabilityConfigurationDAO.delete(observabilityConfig._id)
+        observabilityConfigurationDAO.delete(observabilityConfig._id)
+
+        val setting = observabilityConfig.setting
+        setting.takeIf { it is LangfuseObservabilitySetting }?.let {
+            logger.info { "Deleting the Observability secret ..." }
+            SecurityUtils.deleteSecret((setting as LangfuseObservabilitySetting).secretKey)
+        } ?: logger.info { "No secret to delete for the current setting." }
```
Benvii, line 75: To have a more generic approach using kotlin's reflection APIs, something like this : 

```kotlin
import kotlin.reflect.full.memberProperties
import ai.tock.shared.security.key.SecretKey

fun getSecretKeyAttributes(data: Any): List<SecretKey> {
    return data::class.memberProperties
        .filter { it.returnType.classifier == SecretKey::class } // Filter properties of type SecretKey
        .mapNotNull { it.getter.call(data) as? SecretKey } // Safely get the value as SecretKey
}
```

And apply the deletion on it ... it could be some kind of helper used in all setting ?

bot/admin/server/src/main/kotlin/BotAdminService.kt
Benvii: Good point @AurelienSylvan I agree that it would be cleaner to make a rollback, the deletion fail is logged I think, if we have a good monitoring of warning / errors logs it enough for me.

Benvii (approved):
Overall PR is ok, but I would like your opinion about having a more generic approach for secret deletion using reflection APIs see this conversation here :  TODO

bot/admin/server/src/main/kotlin/service/ObservabilityService.kt
Benvii: Ok for this implementation, can you make this change ?

Benvii (changes requested):
See https://github.com/theopenconversationkit/tock/pull/1762#discussion_r1847957667

Benvii (approved):
Thanks for the last fix, OK for me 👍️

rkuffer (approved):

PR #1764, opened 2024-10-10
resolves #1763 web_connector: add windowFeatures parameter to URL button

### LINKED ISSUE #1763: [Web Connector] Ability to open a popup through URL buttons
The web connector offers the ability to send a link to an arbitrary URL through an `UrlButton`:
https://github.com/theopenconversationkit/tock/blob/903e3c242dd1e43d46ae0499733e6818a2958ffc/bot/connector-web-model/src/main/kotlin/ai/tock/bot/connector/web/send/UrlButton.kt#L22-L29

The existing `target` attribute can be used to make the button either open the link in a new tab (the default), or open in the current tab (specifically, it reflects the [`target` attribute of the HTML anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target)).

For some use cases, notably authentication flows, it can be helpful to instead open a popup window that will then close automatically. Currently, such a popup-opening button has to be implemented through a [widget](https://github.com/theopenconversationkit/tock/tree/master/bot/connector-web#react-chat-widget), but this requires copying existing behaviour from various frontend components.

## Proposed solution

We could add a parameter to `UrlButton` which would enable the popup behaviour. In line with the existing `target` parameter, this new parameter could be a simple string matching the spec of [`windowFeatures`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#windowfeatures). When this `windowFeatures` parameter gets specified, the tock-vue-kit and tock-react-kit would add a JS handler that cancels the default browser behaviour and opens the link in a new window. The implementation of this handler could be as simple as calling `window.open(button.url, button.target, button.windowFeatures)`.

Thoughts @rkuffer @Benvii @assouktim?

Fabilin:
resolves #1763 by adding a new `windowFeatures` parameter to the `UrlButton` interface.

> [!NOTE]
> This new parameter will require an update of the tock-react-kit and tock-vue-kit to be effective.

I did not add the `windowFeatures` argument to the two overloads of `webUrlButton` that take an `HrefTargetType` enum, as it would seem contradictory to try to open a popup window with a predefined `HrefTargetType` that is not `_blank` (the default).
The `@JvmOverloads` annotation is here to preserve binary compatibility with previous versions of the web connector.

Associated PRs:
- theopenconversationkit/tock-react-kit#171

### DIFF
--- a/bot/connector-web-model/src/main/kotlin/ai/tock/bot/connector/web/send/UrlButton.kt
+++ b/bot/connector-web-model/src/main/kotlin/ai/tock/bot/connector/web/send/UrlButton.kt
@@ -25,5 +25,6 @@ data class UrlButton(
     val url: String,
     val imageUrl: String? = null,
     val target: String? = HrefTargetType._blank.name,
-    val style: String? = ButtonStyle.primary.name
+    val style: String? = ButtonStyle.primary.name,
+    val windowFeatures: String? = null,
 ) : Button(ButtonType.web_url)
--- a/bot/connector-web/src/main/kotlin/WebBuilders.kt
+++ b/bot/connector-web/src/main/kotlin/WebBuilders.kt
@@ -98,7 +98,13 @@ fun <T : Bus<T>> T.webButton(
     )
 
 /**
- * Creates a url button
+ * Creates a URL button
+ *
+ * @param title the text that should appear on the button
+ * @param url the URL of the page that should be opened when clicking on the button
+ * @param imageUrl the URL of an image file to display as an icon inside the button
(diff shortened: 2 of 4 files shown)


asus007 (approved):

rkuffer:
Today, it's not possible to build a story offering a `UrlButton` with `windowFeatures` from the studio (as far as I know, the only way to return a `UrlButton` from a simple story is to use the actions of a media card, but neither the front nor the back has been modified to accommodate the new option). 
So, is it possible to return urlButtons from a scripted story? If so, can anyone give me an example of such a story? 
Thanks in advance.

@Fabilin ?

PR #1765, opened 2024-10-14
[Tock Studio] Export Faqs in csv format

rkuffer:
In addition to the json format, Faqs can be exported in csv format. In this case, you can choose the line delimiter and value delimiter for lists, and select the exported columns and their indexes.

faqs-export-csv


AurelienSylvan (approved):

assouktim (approved):
Thanks!

PR #1780, opened 2024-10-30
[Tock Studio] Handle namespaces in state service + refresh dialogs on app change + misc

rkuffer:
- Namespaces retrieval management is delegated to the state service. In this way, the namespace list selector in the application header is updated when a new namespace is created.
- Some competing requests made during application initialization are avoided
- List of dialogs in Analytics > Dialogs view are refereshed when configuration change
- Make sure that `takeUntil` is always the last operator of subscription pipes through the application
- Poorly managed subscriptions corrected on certain components (work to be extended to the entire app)
- Correction of a case not correctly handled in the destruction of the Sentence-training component
- Prevent spinners from being displayed over the application header

### DIFF
--- a/bot/admin/web/src/app/analytics/activity/activity.component.ts
+++ b/bot/admin/web/src/app/analytics/activity/activity.component.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { Component, OnInit } from '@angular/core';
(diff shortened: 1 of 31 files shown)


assouktim (approved):
Thanks a lot!

Benvii (approved):
Thanks 👍️

PR #1784, opened 2024-11-06
[DERCBOT-1193] Fix logging

assouktim:
the deleted failure catch does not exclusively handle authentication errors. Since it is associated with the via (“/*”) route, it will catch all errors generated in this route, including those not strictly related to authentication. This means it can also intercept application errors, network errors or other types of failure. 
(Rollback of https://github.com/CreditMutuelArkea/tock/pull/355)


Benvii (approved):
Thanks for this fix 👍️

rkuffer (approved):

PR #1791, opened 2024-11-21
[DERCBOT-1249] Get a single dialog

assouktim:


assouktim:
The end point already exists

PR #1793, opened 2024-11-22
[GEN AI] Improving Langfuse's traces

assouktim:
Add dialogs details to Langfuse trace & Return Langfuse trace (id, name, url) in RAG Response

### DIFF
--- a/gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/factories/langchain_factory.py
+++ b/gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/factories/langchain_factory.py
@@ -23,7 +23,7 @@ It manages the creation of :
 """
 
 import logging
-from typing import Optional
+from typing import Optional, Any
 
 from langchain_core.embeddings import Embeddings
 from langfuse.callback import CallbackHandler as LangfuseCallbackHandler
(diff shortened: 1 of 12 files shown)


gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/factories/langchain_factory.py
```diff
@@ -338,22 +338,21 @@ def get_callback_handler_factory(
...
 
     Returns:
         The Observability Callback Handler
     """
     if observability_setting is not None:
         return get_callback_handler_factory(
-            setting=observability_setting
-        ).get_callback_handler(trace_name=trace_name.value)
+            setting=observability_setting,
+        ).get_callback_handler(**kwargs)
```
Benvii, line 355: Those are directly mapped to the langfuse attributes ... I'm thinking that this is ok untill we just have langfuse but what about langsmith or Arize Phoenix so I made some research in case we would someday implement them : 

- Arize Phoenix uses auto-instrumentation (compatible with OpenTelemetry standard and OpenInference), meaning it's not done though an explicit langchain callback
- Langsmith SDK is now called LangChainTracer, it seems that they aren't using callbacks any more : context managers are use in the document, tags are passed by wrapping the LCEL chain ...

It seems that langfuse also recommends using decorators : https://langfuse.com/docs/sdk/python/decorators#decorator-context 

I think that it would be great to investigate a bit more and maybe instead of using a callback, wrap chains calls inside python's context managers ... where the manager is build by a factory.
Also we should test LCEL chain tags [with_configs](https://docs.smith.langchain.com/observability/how_to_guides/tracing/trace_with_langchain#add-metadata-and-tags-to-traces) (to pass tags) in a more generic way, maybe we would have to pass them to the observability provider.

Benvii (changes requested):
Thanks for this PR, ok for me juste added some suggestions and reflexions about our callback integration approach, if you agree I will create an other ticket for that I thing our current observability integration needs to be updated.

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/factories/langchain_factory.py
assouktim: yes, it's a good thing, we need to move towards this new way of tracing. As seen together on a separate ticket.

Benvii: Created on #1819

Benvii (approved):
Thanks for the updates ok for me.

Benvii:
I'm rebasing it and fixing the conflits and I will merge it.

scezen (approved):

Benvii (approved):

PR #1798, opened 2024-11-29
[Tock Studio] Add the ability to open a single dialog in a view with url

rkuffer:
- Adds the ability to display logs for a single dialog in a view with a single url
- Sets the studio to the correct namespace/application pair when the url of a dialog log view is called
- Scroll and visually identify the targeted message (conveyed by url fragment)
- Graphical refactoring of dialog log display + creation of a dedicated component

image


assouktim (approved):
Thanks a lot!

Benvii (approved):
Thanks 👍️

Benvii:
Merging it as there is no backend dependencies

PR #1805, opened 2024-12-09
[Tock Studio] Fix answer label namespace case

rkuffer:
During creation, the backend sets the namespace of answers labels to lowercase. This produces a bug when updating the label of a response whose namespace contains uppercase characters. This PR bypasses this bug until the backend behavior is corrected.


assouktim (approved):
Thanks !

Benvii (approved):
Thanks for this fix 👍️

PR #1806, opened 2024-12-10
Fix - Document index name generation when no vector store has been defined

assouktim:


Benvii (approved):
Thanks for this fix I understand that VectorStoreUtils is based on environment variables so default to the correct index normalization.

scezen (approved):

PR #1820, opened 2025-01-10
[DERCBOT-1327] Git hooks - Gitleaks

assouktim:
Git hook to detect secrets such as passwords, API keys and tokens in git repositories.

## Breaking change
For deployment using RAW secrets stored in mongo database or calling the Gen AI Orchestrator using raw secret (which is not advised for obvious security reasons), the format of the secret is changed key containing the secret value was previously named api_key.value is now named api_key.secret.

NB : We have not updated the existing data, as we consider that only developers use Raw-type secrets (which are subject to code refactoring)


gen-ai/orchestrator-server/src/main/python/server/.pre-commit-config.yaml
Benvii, line 1: Is there an advantage to move it at the repository base ? or is it because pre-commit doesn't support multiple config files ?

Benvii (changes requested):
Thanks for this PR 👍️

Missing :
- [x] Config settings and ragquery exemples here : https://github.com/theopenconversationkit/tock/tree/master/gen-ai/orchestrator-server/src/main/python/tock-llm-indexing-tools/examples 

Can you also add to the main README.md how to setup this hook / pre-commit, so that new developers will have the documentation.

gen-ai/orchestrator-server/src/main/python/server/.pre-commit-config.yaml
assouktim: Yes the pre-commit doesn't support multiple config files.

Benvii (approved):

Benvii (changes requested):
Sorry I forgot one point, can you add to the main README.md that this project uses pre-commit ?

I agree that this README.md should have more content, we can start with this :)
Otherwise developers will forgot to enable the pre-commit hook.

Benvii (approved):
Thanks 👍️

rkuffer (approved):

PR #1825, opened 2025-01-16
Increase test dialog modal max height

rkuffer:
Increasing the maximum possible height of the dialog test window


Benvii (approved):
Thanks for this update 👍️

scezen (approved):

PR #1835, opened 2025-01-28
[DERCBOT-1385] Do not retrieve “Inboxed” phrases with FAQ questions.

assouktim:
This PR ensure that the FAQ question tab only gets the sentences that were train for this FAQ and not all the sentences classified automatically for this FAQ.

### DIFF
--- a/nlp/front/storage-mongo/src/main/kotlin/FaqDefinitionMongoDAO.kt
+++ b/nlp/front/storage-mongo/src/main/kotlin/FaqDefinitionMongoDAO.kt
@@ -298,7 +298,7 @@ object FaqDefinitionMongoDAO : FaqDefinitionDAO {
                 // join FaqDefinition with IntentDefinition
                 joinOnIntentDefinition(),
                 // join FaqDefinition with ClassifiedSentence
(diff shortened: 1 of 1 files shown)


nlp/front/storage-mongo/src/main/kotlin/FaqDefinitionMongoDAO.kt
```diff
@@ -486,6 +486,10 @@ object FaqDefinitionMongoDAO : FaqDefinitionDAO {
                                 ne from listOf(
                                     ClassifiedSentence::status,
                                     ClassifiedSentenceStatus.deleted
+                                ),
+                                ne from listOf(
+                                    ClassifiedSentence::status,
+                                    ClassifiedSentenceStatus.inbox
```
Benvii, line 492: Why not explicitly getting validated and model ? I think it would be easier for someone who reads the code.

And it will be more stable if we introduce new status in the futur (archived ...)

Benvii (changes requested):
Ok b

Benvii (approved):
Ok for me 👍️

scezen (approved):

PR #1846, opened 2025-02-12
[DERCBOT-1394] Web Connector - Add web security mode

assouktim:


shared/src/main/kotlin/security/auth/spi/WebSecurityHandler.kt
Fabilin, line 1: Shouldn't this be in the web connector subproject?

bot/connector-web/src/main/kotlin/WebConnectorProvider.kt
```diff
@@ -16,32 +16,50 @@
...
-import ai.tock.shared.resourceAsString
+import ai.tock.bot.connector.*
+import ai.tock.shared.*
+import ai.tock.shared.security.auth.spi.WebSecurityHandler
+import ai.tock.shared.security.auth.spi.WebSecurityMode
 import kotlin.reflect.KClass
 
+private const val WEB_SECURITY_MODE_PARAM = "web_security_mode"
+
+val cookieAuth = booleanProperty("tock_web_cookie_auth", false)
```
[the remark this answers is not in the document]
assouktim: This method was public, as it is also used in the SSE routes that I hadn't impacted. Now the SSE also uses WebSecurityMode, so I've made it private.

bot/connector-web/src/main/kotlin/security/WebSecurityPassthroughHandler.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import ai.tock.shared.security.auth.spi.WebSecurityHandler
+import io.vertx.ext.web.RoutingContext
+import mu.KotlinLogging
+
+class WebSecurityPassthroughHandler : WebSecurityHandler {
+    private val logger = KotlinLogging.logger {}
+
+    override fun handle(routingContext: RoutingContext) {
+        logger.info { "Nothing to do." }
```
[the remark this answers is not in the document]
assouktim: No need in fact, I've put the liveau log in debug mode

shared/src/main/kotlin/security/auth/spi/WebSecurityHandler.kt
assouktim: No, you need to keep it outside, as it can be used on other connectors.

bot/connector-web/src/main/kotlin/WebConnectorProvider.kt
```diff
@@ -16,32 +16,50 @@
...
+                // But if it's missing or set to "DEFAULT" it chooses between two options :
+                // If "cookieAuth" is enabled, it picks COOKIES.
+                // If not, it picks PASSTHROUGH (which likely means no special security measures).
+                ?: if (cookieAuth) WebSecurityMode.COOKIES else WebSecurityMode.PASSTHROUGH
+
             return WebConnector(
                 connectorId,
-                path
+                path,
+                injector.provide<WebSecurityHandler>(tag = webSecurityType.name)
```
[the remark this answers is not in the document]
assouktim: The ServiceLoader is used to load/create instances, which is done using the mechanism we already have (BotAdditionalModulesService). Here, I retrieve an instance that has already been loaded, and the tag allows me to dynamically choose the right implementation based on the bot configuration. The use of KodeIn here is quite intentional.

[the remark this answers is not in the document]
Fabilin: That's perfectly fair

bot/connector-web/README.md
```diff
@@ -251,3 +251,27 @@ This connector can process [Markdown formatting](https://daringfireball.net/proj
...
+### 3. `PASSTHROUGH`
+- This mode allows all requests to pass through without any security enforcement.
+- No authentication or authorization checks are performed.
+
+### 4. `JWT`
+- This mode validates JSON Web Tokens (JWT) included in requests.
+- It checks the signature, ensures the token has not been revoked, and manages user authorization.
+- This mode is recommended for stateless authentication in secure environments.
+- The implementation of JWT validation is not provided by default, as each company has its own approach to verifying tokens.
+- To use this mode, you must provide an implementation using Java's Service Loader mechanism.
```
Fabilin, line 276: I believe it would be helpful to detail this part a bit more, here's a suggestion:
```suggestion
- To use this mode, you must provide an implementation of `BotAdditionalModulesService` using Java's Service Loader mechanism, and use it to bind your implementation of `WebSecurityHandler` with the tag "OAUTH"
```

shared/src/main/kotlin/security/auth/spi/WebSecurityMode.kt
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.tock.shared.security.auth.spi
+
+enum class WebSecurityMode {
```
Fabilin, line 19: Actually, given that the input in TOCK Studio is a string (I don't think we can replace it with a `select` at the moment), and that anyone can provide their own implementation with a custom binding, does it still make sense to constrain the possible values to just these four?

I could imagine multiple JWT providers being present, and multiple web connectors each using their own.

assouktim: Yes, I agree with you, but today only one use case has been identified for this authentication, so I suggest you leave it as it is, and consider opening it up further in the future if other implementations come to be added.

Fabilin (approved):

vsct-jburet (approved):

PR #1849, opened 2025-02-17
[DERCBOT-1365] Improving web scraping

assouktim:


Benvii (approved):
Thanks for adding this  min parameter and this code refactoring also 👍️

rkuffer (approved):

PR #1854, opened 2025-02-24
resolves #1853 Whatsapp Cloud : Add phone number and username to UserPreferences

### LINKED ISSUE #1853: [WhatsApp Cloud] Add user informations in UserPreferences
In some case we need to get the user name or phone number, for instance to share it with customer service team.
Add the ability to retrieve retrieve them in userPreferences

gsoumier:


bot/engine/src/main/kotlin/engine/user/UserPreferences.kt
```diff
@@ -60,7 +60,16 @@ data class UserPreferences(
     /**
      * The initial locale of the user
      */
-    var initialLocale: Locale = locale
+    var initialLocale: Locale = locale,
+    /**
+     * Full name or screen name of the user in the application.
+     */
+    var userName: String? = null,
```
Fabilin, line 67:
```suggestion
    var username: String? = null,
```
All lowercase seems more standard

Fabilin (approved):
Future improvement: add username property to other connectors (e.g. google assistant)

Also, we should provide a mechanism to avoid saving personal information while a user hasn't accepted a bot's legal mentions (separate issue?).

vsct-jburet (approved):

PR #1858, opened 2025-03-04
:bug: feat(llm-indexing-tools) : Fix requirements.txt local path dependency

Benvii:
Fix requirements.txt local path dependency exported into the zip dependency.

This leads to errors when build docker images.


assouktim (approved):
Thanks!

PR #1856, opened 2025-02-26
[DERCBOT-1374] Public URL for Langfuse Observability Settings

scezen:
Ticket : [DERCBOT-1374](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1374)

This PR introduces a new optionnal attribute for the Observability Settings :

```
data class LangfuseObservabilitySetting<T>(
    override val secretKey: T,
    val publicKey: String,
    val url: String,
    val publicUrl : String? = null,
) : ObservabilitySettingBase<T>(ObservabilityProvider.Langfuse), HasSecretKey<T>
```

If filled, the traceUrl of `ObservabilityInfo` will get replaced by the public URL as so : 

```json
"botMetadata": {
"lastAnswer": false,
"priority": "normal",
"visibility": "UNKNOWN",
"quoteMessage": "UNKNOWN",
"orchestrationLock": false,
"orchestratedBy": null,
"returnsHistory": false,
"debugEnabled": false,
"sourceWithContent": false,
"isGenAiRagAnswer": true,
"streamedResponse": false,
"observabilityInfo": {
"traceId": "6832661a-ed85-404d-abff-71a1f50bf410",
"traceName": "RAG",
"traceUrl": "http://langfuse-public:3000//trace/6832661a-ed85-404d-abff-71a1f50bf410"
}
```

PR #1873, opened 2025-03-24
[Tock Studio] Add of observability public url input

rkuffer:
Works with https://github.com/theopenconversationkit/tock/pull/1856

In Gen AI > Observability, define an optional public access url to the observability tool


Benvii (approved):
Ok, just waiting for a small fix on https://github.com/theopenconversationkit/tock/pull/1856 before merging it.

PR #1874, opened 2025-03-26
[Tock Studio] Warning on inbox faq sentences + link to sentences search view

rkuffer:
- Added a closable warning to the “Question” tab of the faq editing view to inform the user that only “validated” and “in model” sentences are listed.
- The warning includes a link to the “Language understanding > Search sentences” view. If followed, the link opens the view with the advanced search field “Intent” filled in with the faq's intent and the “Inbox” option checked, and launches the search.

image


Benvii (approved):
Thanks 👍️

PR #1877, opened 2025-03-28
fix: correct variable name and class reference in RAG Chain

scezen:
This pull request focuses on refactoring the `RAGChain` service in the `gen-ai/orchestrator-server` to use the `request` object instead of the `query` object. This change aims to standardize the codebase and improve consistency in the handling of request data.

Key changes include:

* **Refactoring to use `request` object:**
  * Updated `create_rag_chain` method to validate prompt templates using `request` instead of `query`.
  * Modified `get_rag_debug_data` method to retrieve dialog history and user question from `request` instead of `query`.


Benvii (approved):
Thanks 👍️

assouktim (approved):
Thanks!

PR #1889, opened 2025-05-14
[DERCBOT-1500] Allow namespace application listing without tock_namespace_open_access

scezen:
[DERCBOT-1500](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1500)

Fix namespace application listing endpoint to allow access if user has namespace rights, without relying solely on tock_namespace_open_access


Benvii (approved):
Thanks

vsct-jburet (approved):

PR #1904, opened 2025-07-03
Fix double flush when switching to no-answer story (proactive handler)

scezen:
[DERCBOT-1531](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1531)

### Problem
Prevent a second flush/HTTP write when RAG returns NO_RAG_SENTENCE and the flow switches to the configured no-answer story.

### Fix
* Skip flush/close in AbstractProactiveAnswerHandler when a fallback
  story is returned.
* Force PROACTIVE_CONVERSATION_STATUS = CLOSED before calling
  handleAndSwitchStory(), so the convo is not closed a second time.
* Dialog is persisted after the new story has added its answer.


Benvii (approved):
LGTM thanks.

rkuffer (approved):

PR #1908, opened 2025-08-13
resolves #1907 bot_engine: implement coroutine-based stories

### LINKED ISSUE #1907: [Bot Engine] allow asynchronous operations in story handlers
Currently, stories in TOCK must answer synchronously, blocking a thread until completion. However, more and more story handler implementations rely on long-running operations, like API calls and LLM routines. As such, it would be useful to have the possibility of writing an asynchronous story handler, which could perform asynchronous operations before sending a message.

> [!IMPORTANT]
> TOCK's existing concurrency management relies on Vert.x to manage threads. To keep backward compatibility and avoid creating too many concurrent threads, the new feature should be based on the existing [`Executor`](https://github.com/theopenconversationkit/tock/blob/e25e3a124795481a60ad2b17e5473cf5853ecc20/shared/src/main/kotlin/Executor.kt) interface.

## Possible concurrency APIs
We have a few options to manage concurrency in async stories:

1. Vert.x tasks: naturally supported, but further exposes Vert.x in our public API. Developers may not be familiar with this API.
2. `CompletableFuture`: already supported in `Executor`. As part of the JDK, developers are likely to be familiar with this API.
3. Kotlin coroutines: can be bridged from `Executor` using [`asCoroutineDispatcher`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/as-coroutine-dispatcher.html). Kotlin developers are likely to be familiar with this API, and it is the only one that provides structured concurrency.

I believe we should aim to provide support for Kotlin coroutines, as it is the modern standard for Kotlin applications. What are your thoughts?

Fabilin:
resolves #1907 by adding new subtypes of story handlers with `suspend` methods. These methods are called by the engine based on an instance check.

### DIFF
--- a/bot/engine/src/main/kotlin/engine/Bot.kt
+++ b/bot/engine/src/main/kotlin/engine/Bot.kt
@@ -33,11 +33,12 @@ import ai.tock.bot.engine.dialog.Story
 import ai.tock.bot.engine.feature.DefaultFeatureType
 import ai.tock.bot.engine.nlp.NlpController
 import ai.tock.bot.engine.user.UserTimeline
+import ai.tock.shared.coroutines.ExperimentalTockCoroutines
 import ai.tock.shared.injector
 import com.github.salomonbrys.kodein.instance
 import java.util.Locale
+import kotlinx.coroutines.runBlocking
 import mu.KotlinLogging
-import kotlin.time.Duration
 
 /**
  *
@@ -81,6 +82,14 @@ internal class Bot(botDefinitionBase: BotDefinition, val configuration: BotAppli
      * Handle the user action.
      */
     fun handle(action: Action, userTimeline: UserTimeline, connector: ConnectorController, connectorData: ConnectorData) {
+        @OptIn(ExperimentalTockCoroutines::class)
+        runBlocking {
+            handleAction(action, userTimeline, connector, connectorData)
+        }
+    }
+
+    @ExperimentalTockCoroutines
+    suspend fun handleAction(action: Action, userTimeline: UserTimeline, connector: ConnectorController, connectorData: ConnectorData) {
         connector as TockConnectorController
 
         loadProfileIfNotSet(connectorData, action, userTimeline, connector)
@@ -108,7 +117,7 @@ internal class Bot(botDefinitionBase: BotDefinition, val configuration: BotAppli
 
         if (!userTimeline.userState.botDisabled) {
             dialog.state.currentIntent?.let { intent ->
-                connector.sendIntent(intent, action.applicationId, connectorData)
+                connector.sendIntent(intent, action.connectorId, connectorData)
             }
             connector.startTypingInAnswerTo(action, connectorData)
             val story = getStory(userTimeline, dialog, action)
@@ -122,7 +131,7 @@ internal class Bot(botDefinitionBase: BotDefinition, val configuration: BotAppli
 
             try {
                 currentBus.set(bus)
-                story.handle(bus)
+                story.handleAsync(bus)
                 if (shouldRespondBeforeDisabling) {
                     userTimeline.userState.botDisabled = true
                 }
--- a/bot/engine/src/main/kotlin/engine/TockConnectorController.kt
+++ b/bot/engine/src/main/kotlin/engine/TockConnectorController.kt
@@ -27,8 +27,8 @@ import ai.tock.bot.engine.action.SendAttachment
 import ai.tock.bot.engine.action.SendAttachment.AttachmentType.audio
 import ai.tock.bot.engine.action.SendSentence
 import ai.tock.bot.engine.event.Event
-import ai.tock.bot.engine.event.TypingOnEvent
 import ai.tock.bot.engine.event.MetadataEvent
+import ai.tock.bot.engine.event.TypingOnEvent
 import ai.tock.bot.engine.user.PlayerId
 import ai.tock.bot.engine.user.UserLock
 import ai.tock.bot.engine.user.UserPreferences
@@ -36,6 +36,8 @@ import ai.tock.bot.engine.user.UserTimeline
 import ai.tock.bot.engine.user.UserTimelineDAO
 import ai.tock.shared.Executor
 import ai.tock.shared.booleanProperty
+import ai.tock.shared.coroutines.ExperimentalTockCoroutines
+import ai.tock.shared.coroutines.launchCoroutine
 import ai.tock.shared.error
 import ai.tock.shared.injector
 import ai.tock.shared.intProperty
@@ -44,10 +46,10 @@ import ai.tock.shared.provide
 import ai.tock.stt.STT
 import com.github.salomonbrys.kodein.instance
 import io.vertx.ext.web.Router
-import mu.KotlinLogging
 import java.net.URL
-import java.time.Duration
 import java.util.concurrent.CopyOnWriteArrayList
+import kotlinx.coroutines.delay
+import mu.KotlinLogging
 
 private val synchronousMode = booleanProperty("tock_timeline_persistence_synchronous_mode", true)
 private val asynchronousMode = booleanProperty("tock_timeline_persistence_asynchronous_mode", false)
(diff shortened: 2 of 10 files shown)


bot/engine/src/main/kotlin/engine/AsyncBotBus.kt
```diff
@@ -0,0 +1,296 @@
...
+        set(step) {
+            synchronized(botBus) {
+                story.step = step?.name
+            }
+        }
+    override val userInfo: UserPreferences
+        get() = botBus.userPreferences
+    override val userState: UserState
+        get() = botBus.userTimeline.userState
+    override var nextUserActionState: NextUserActionState?
```
Fabilin: could replace with a `suspend fun constrainNlp(nextState: NextUserActionState?)`

bot/engine/src/main/kotlin/engine/BotBus.kt
```diff
@@ -462,6 +475,12 @@ interface BotBus : Bus<BotBus> {
         storyDefinition.storyHandler.handle(this)
     }
 
+    @Deprecated("Do not switch to an AsyncStoryDefinition from a synchronous story")
```
Fabilin: Can be marked as error directly

bot/engine/src/main/kotlin/engine/CoroutineBridgeBus.kt
```diff
@@ -0,0 +1,135 @@
...
+        switchStory(storyDefinition, starterIntent)
+        hasCurrentSwitchStoryProcess = false
+        val storyHandler = storyDefinition.storyHandler
+
+        if (!handleAsyncStory(storyHandler, currentStoryId)) {
+            storyHandler.handle(this)
+        }
+    }
+
+    fun handleAsyncStory(callingStoryId: String, op: suspend (AsyncBotBus) -> Unit): Boolean {
```
Fabilin: Think about removing this and disabling handleAndSwitchStory to async

bot/engine/src/main/kotlin/definition/StepExecutionContext.kt
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package ai.tock.bot.definition
+
+/**
+ * @see StoryHandlerDefinition
+ * @see AsyncStoryHandling
+ */
+interface StepExecutionContext
```
Fabilin: can be removed now

bot/engine/src/main/kotlin/engine/TockBotBus.kt
```diff
@@ -222,4 +240,29 @@ internal class TockBotBus(
...
         }
     }
+
+    /**
+     * @return a callback to force-close the message queue
+     */
+    fun deferMessageSending(scope: CoroutineScope): () -> Unit {
+        data class QueuedAction(val action: Action, val delay: Long)
+
+        val messageChannel = Channel<QueuedAction>(Channel.BUFFERED)
```
vsct-jburet, line 250: What of Channel is not closed ?

bot/engine/src/main/kotlin/engine/user/UserLock.kt
```diff
@@ -23,5 +23,16 @@ interface UserLock {
...
 
+    /**
+     * Acquires the user lock only if it is free at the time of invocation
+     *
+     * Acquires the lock for the given [userId] if it is available and returns immediately
+     * with the value `true`.
+     * If the lock is not available then this method will return
+     * immediately with the value `false`.
+     */
+    // TODO implement using mongo-coroutines
```
vsct-jburet: please remove the TODO or implement it ;)

bot/engine/src/main/kotlin/engine/AsyncBotBus.kt
```diff
@@ -0,0 +1,270 @@
...
+        withContext(executor.asCoroutineDispatcher()) {
+            synchronized(botBus) {
+                botBus.send(i18nText, delay = delay)
+            }
+        }
+    }
+
+    override suspend fun send(i18nText: String, vararg i18nArgs: Any?) {
+        withContext(executor.asCoroutineDispatcher()) {
+            synchronized(botBus) {
```
vsct-jburet, line 214: Do you keep the synchronized ?

bot/engine/src/main/kotlin/engine/Bot.kt
```diff
@@ -108,26 +111,28 @@ internal class Bot(botDefinitionBase: BotDefinition, val configuration: BotAppli
...
+                    story.handle(asyncBus)
+                    if (shouldRespondBeforeDisabling) {
+                        userTimeline.userState.botDisabled = true
+                    }
                 }
-            } finally {
-                currentBus.remove()
+
+                // Ensure we do not have a lingering message sending job
+                closeMessageQueue()
```
vsct-jburet, line 135: May be a a try finally is better here

bot/engine/src/main/kotlin/engine/TockConnectorController.kt
```diff
@@ -143,14 +157,15 @@ internal class TockConnectorController(
...
+    @ExperimentalTockCoroutines
+    private suspend fun handleAction(action: Action, nbAttempts: Int, data: ConnectorData) {
         val callback = data.callback
         try {
             val playerId = action.playerId
             val id = playerId.id
 
             logger.debug { "try to lock $playerId" }
-            if (userLock.lock(id)) {
+            if (userLock.tryLock(id)) {
```
vsct-jburet, line 168: In this current state, here you make a blocking call in a suspend block :(

bot/engine/src/main/kotlin/engine/Bot.kt
```diff
@@ -80,7 +82,8 @@ internal class Bot(botDefinitionBase: BotDefinition, val configuration: BotAppli
     /**
      * Handle the user action.
      */
-    fun handle(action: Action, userTimeline: UserTimeline, connector: ConnectorController, connectorData: ConnectorData) {
+    @ExperimentalTockCoroutines
+    suspend fun handleAction(action: Action, userTimeline: UserTimeline, connector: ConnectorController, connectorData: ConnectorData) {
```
vsct-jburet, line 86: the code in this method has to be non blocking - it isn't

bot/engine/src/main/kotlin/engine/TockBotBus.kt
Fabilin: If we receive no "last message", the channel will be closed at the end of `Bot.handleAction`. `TockBotBus` is an internal class, so we don't have to worry about third-party code calling this method and forgetting to close the Channel.

bot/engine/src/main/kotlin/engine/AsyncBotBus.kt
Fabilin: I think it's important to keep it here, considering `TockBotBus.answer` has a lot of state mutation that must happen atomically.

bot/engine/src/main/kotlin/engine/Bot.kt
Fabilin: Is it necessary actually? If an exception is thrown, the coroutine scope will be cancelled, in turn closing the channel.

bot/engine/src/main/kotlin/engine/TockConnectorController.kt
Fabilin: Correct, sad emoji. Will have to be fixed in a MongoDB coroutine PR later. Should I leave a comment in the code?

bot/engine/src/main/kotlin/engine/Bot.kt
Fabilin: I believe the main blocking code left is the database calls, which should be fixed in a later PR. Is there anything else?

bot/engine/src/main/kotlin/engine/Bot.kt
Fabilin: I was somewhat confused, there is no special interaction between cancellation and channels, but it should nonetheless be fine: https://stackoverflow.com/questions/43889066/what-closing-a-kotlinx-coroutines-channel-does

vsct-jburet (approved):

PR #1913, opened 2025-09-04
[DERCBOT-1609] Improving RAG System - Part 1

assouktim:
In this first part of improvement of the RAG system, we proceeded as follows:
- Structuring the LLM response to obtain not only the answer, but also other information such as: the status of the response, the subject of the conversation, and the actual context used.
- Modifying the default prompt for condensing questions.
- Preparing the code for the multiple search stage.


bot/admin/web/src/app/rag/rag-settings/models/engines-configurations.ts
```diff
@@ -27,7 +27,19 @@ import {
...
+
+Your task:
+- Reformulate the user’s latest question into a clear, standalone query.
+- Incorporate relevant context from the conversation history.
+- Do NOT answer the question.
+- If the history does not provide additional context, keep the question as is.
+
+Return only the reformulated question.`;
 
 export const QuestionAnsweringDefaultPrompt: string = `# TOCK (The Open Conversation Kit) chatbot
```
Benvii: As JsonOutputParser.get_format_instructions() isn't used in the Rag Chain to include instructions (in the question answrering prompt) about the requested output format (like in langchain exemple [here](https://python.langchain.com/docs/how_to/output_parser_json/#streaming:~:text=we%20are%20passing-,format_instructions,-from%20the%20parser)) you need to update the default template to include the formatting instruction part otherwise users can't guess it.

Here is the prompt saved in our internal ticketing system : 
```
## Instructions:
You must answer STRICTLY in valid JSON format (no extra text, no explanations).
Use only the following context and the rules below to answer the question.

### Rules for JSON output:
- If the answer is found in the context:
  - "answer": "<the best possible answer in {locale}>"
  - "status": "found_on_context"

- If the answer is NOT found in the context:
  - "answer": null
  - "status": "not_in_context"

- If the question matches a special case defined below:
  - "answer": null
  - "status": "<the corresponding case code>"

And for all cases (MANDATORY):
  - "topic": "<exactly ONE topic chosen STRICTLY from the predefined list below. If no exact match is possible, set 'unknown'>"
  - "suggested_topics": ["<zero or more free-form suggestions if topic is unknown>"]

### Special Cases:
<case: code>
(Example: if the question is about account closure â†’ status = "redirect_to_human")

### Predefined list of topics (use EXACT spelling, no variations):
TODO ... list your topics.

## Context:
{context}

## Question:
{question}

## Output format (JSON only):
Return your response in the following format:

```json
{{
  "answer": "TEXTUAL_ANSWER_OR_NULL",
  "status": "found_on_context" | "not_in_context" | "redirect_to_human" | "OTHER_CASE_CODE",
  "topic": "EXACT_TOPIC_FROM_LIST_OR_UNKNOWN",
  "suggested_topics": [
    "SUGGESTED_TOPIC_1",
    "SUGGESTED_TOPIC_2"
  ]
}}
```

If we want to use streaming in the future is would be great to start this answer of with status field maybe so that the partiel JSON as directly this field before streaming the answer field, see this exemple of how it's streamed here : 
https://python.langchain.com/docs/how_to/output_parser_json/#streaming

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/models/rag/rag_models.py
```diff
@@ -52,16 +52,17 @@ class Footnote(Source):
...
-
-    text: str = Field(
-        description='Text with footnotes used to list outside sources',
-        examples=['This is page content [1], and this is more content [2]'],
-    )
-    footnotes: set[Footnote] = Field(description='Set of footnotes')
-
+class ChunkSentences(BaseModel):
+    chunk: Optional[str] = None
+    sentences: Optional[List[str]] = None
```
Benvii: Add a description of what theses sentences are, I assume it's the part of the chunk used to formulate the answer ? Or maybe it's the sentence form the answer related to this chunk ?

It's not describe in the prompt template you gave on JIRA.

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/models/rag/rag_models.py
```diff
@@ -52,16 +52,17 @@ class Footnote(Source):
...
+    reason: Optional[str] = None
```
Benvii: Same question here, what's this reason field used for ?

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/models/rag/rag_models.py
```diff
@@ -52,16 +52,17 @@ class Footnote(Source):
...
-    """Text with its footnotes. Used for RAG response"""
-
-    text: str = Field(
-        description='Text with footnotes used to list outside sources',
-        examples=['This is page content [1], and this is more content [2]'],
-    )
-    footnotes: set[Footnote] = Field(description='Set of footnotes')
-
+class ChunkSentences(BaseModel):
+    chunk: Optional[str] = None
```
Benvii: How are chunk linked to their source / footnotes ? As I understand there is no link between them and we can't link them in the first implementation.

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
```diff
@@ -120,96 +123,109 @@ async def execute_rag_chain(
...
-        user_id = (request.dialog.user_id,)
-        tags = (request.dialog.tags,)
 
-    logger.debug(
-        'RAG chain - Use chat history: %s',
-        'Yes' if len(message_history.messages) > 0 else 'No',
+    logger.debug('RAG chain - Use chat history: %s', len(message_history.messages) > 0)
+    logger.debug('RAG chain - Use RAGCallbackHandler for debugging : %s', debug)
+
+    callback_handlers = get_callback_handlers(request, custom_observability_handler, debug)
```
Benvii: Isn't it easier to return here a Tuple or an object (with 2 fields {records_callback_handler: [], observability_handler: [] }, which explicitly split the 2 kind of handlers that are instanciated in get_callback_handlers, instead of having to split them back in 2 categories RAGCallbackHandler and LangfuseCallbackHandler ?

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
```diff
@@ -120,96 +123,109 @@ async def execute_rag_chain(
...
+    records_callback_handler = None
+    if debug:
+        records_callback_handler = next(
+            (x for x in callback_handlers if isinstance(x, RAGCallbackHandler)),
+            None
+        )
+    observability_handler = next(
+        (x for x in callback_handlers if isinstance(x, LangfuseCallbackHandler)),
```
Benvii: Mentioning langfuse here isn't really generic if we have a arize phoenix handler in the future (which is used by SNCF Connect).

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
```diff
@@ -451,6 +504,8 @@ def get_rag_documents(handler: RAGCallbackHandler) -> List[RAGDocument]:
         for doc in handler.records['documents']
     ]
 
+def get_llm_answer(rag_chain_output) -> LLMAnswer:
+    return LLMAnswer(**json.loads(rag_chain_output.strip().removeprefix("```json").removesuffix("```").strip()))
```
Benvii: Is this manual parsing required ? Langchain doesn't have any parser for that kind of markdown parsing ?

I think you can call this instead : https://python.langchain.com/api_reference/core/output_parsers/langchain_core.output_parsers.json.JsonOutputParser.html#langchain_core.output_parsers.json.JsonOutputParser.parse

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
```diff
@@ -279,31 +295,61 @@ def create_rag_chain(
     if question_condensing_llm_factory is not None:
         question_condensing_llm = question_condensing_llm_factory.get_language_model()
     question_answering_llm = question_answering_llm_factory.get_language_model()
-    rag_prompt = build_rag_prompt(request)
 
-    # Construct the RAG chain using the prompt and LLM,
-    # This chain will consume the documents retrieved by the retriever as input.
-    rag_chain = construct_rag_chain(question_answering_llm, rag_prompt)
+    if question_condensing_llm is not None:
```
Benvii, line 307:
```suggestion
    # Fallback in case of missing condensing LLM setting using the answering LLM setting.
    if question_condensing_llm is not None:
```

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
```diff
@@ -279,31 +295,61 @@ def create_rag_chain(
...
+        "question": itemgetter("condensed_question"),
+        "chat_history": itemgetter("chat_history"),
+        "documents": RunnableLambda(retrieve_with_variants),
+    })
+
+    return rag_inputs | RunnablePassthrough.assign(answer=(
+            {
+                "context": lambda x: json.dumps([
+                    {
+                        "chunk_id": doc.metadata['id'],
```
Benvii: Great to have a chunk_id here but I don't see where the LLM should reuse this chunk_id in it's reply as it's not used / present in the output type, ChunkSentences doesn't have an ID field.

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
```diff
@@ -316,40 +362,59 @@ def build_rag_prompt(request: RAGRequest) -> LangChainPromptTemplate:
...
-    """
     return (
         {
-            'context': lambda inputs: '\n\n'.join(
-                doc.page_content for doc in inputs['documents']
-            ),
-            'question': lambda inputs: inputs[
-                'question'
-            ],  # Override the user's original question with the condensed one
+            "context": lambda x: json.dumps([
```
Benvii: It seems to be code duplicated here, I don't see construct_rag_chain used anywhere in the code base .. how is it different from create_rag_chain ?

I see that create_rag_chain is imported in the tooling run_experiment.py script but it's unused 😬, actually it's execute_rag_chain that is used in the tooling so **construct_rag_chain should be removed**.

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/models/rag/rag_models.py
assouktim: Reason why the chunk was not used (e.g., irrelevant, general background).

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/models/rag/rag_models.py
assouktim: Chunks are linked to footnotes through their ID:
ChunkInfos.chunk matches Document.metadata["id"].   (i renamed ChunkSetences to ChunkInfos)
See the instruction that returns RAGResponse (on rag_chain.py).

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
assouktim: No, it's not the return of LLM, but what the log handler records. It's a string that it sees.
We do use JsonOutputParser in RAG chain.

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain.py
assouktim: We explain that on prompt : 

- If explicit chunk identifiers are present in the context, use them; otherwise assign sequential numbers starting at 1.
- For each chunk object:
  - "chunk": "<chunk_identifier_or_sequential_number>"

bot/admin/web/src/app/rag/rag-settings/models/engines-configurations.ts
```diff
@@ -27,35 +27,102 @@ import {
...
+{{ chat_history }}
+
+## User question
+{{ question }}
+
+## Output format (JSON only):
+Return your response in the following format:
+
+{
+  "status": "found_on_context" | "not_in_context" | "small_talk",
```
scezen: Small typo but should be "found_in_context" I think.

PR #1920, opened 2025-09-29
resolves #1919 use coroutine for critical mongo queries

### LINKED ISSUE #1919: use coroutine for critical mongo queries

vsct-jburet:


bot/engine/src/main/kotlin/engine/feature/FeatureDAO.kt
```diff
@@ -38,7 +38,7 @@ interface FeatureDAO {
     ): Boolean =
         isEnabled(botId, namespace, category, name, null, default, userId)
 
-    fun isEnabled(
+    suspend fun isEnabled(
```
Fabilin, line 41: This change would break a lot of user code, which makes me think a compatibility fix may be necessary. I would suggest having a separate `AsyncFeatureDAO` interface with all the suspending methods, keeping the `FeatureDAO` interface intact, and making its implementation delegate to the former using `runBlocking`.

bot/engine/src/main/kotlin/engine/user/UserLock.kt
```diff
@@ -21,7 +21,7 @@ package ai.tock.bot.engine.user
  */
 interface UserLock {
 
-    fun lock(userId: String): Boolean
+    suspend fun lock(userId: String): Boolean
```
Fabilin, line 24: My base idea here was to keep this method in its blocking form and deprecate it, so we can have a cycle of deprecation before breaking compatibility with code that may use the lock. It may be too advanced a use case to warrant this kind of care though.

bot/storage-mongo/src/main/kotlin/FeatureMongoDAO.kt
```diff
@@ -49,14 +45,18 @@ internal data class Feature(
     val graduation: Int? = null,
 )
 
-internal class FeatureMongoDAO(private val cache: FeatureCache, private val col: MongoCollection<Feature>) :
-    FeatureDAO {
+internal class FeatureMongoDAO(
+    private val cache: FeatureCache,
+    asyncCol: com.mongodb.reactivestreams.client.MongoCollection<Feature>,
```
Fabilin: Missing import?

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -261,16 +326,16 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
             }
         }
 
-        if(asynchronousProcess) {
-            executor.executeBlocking { saveProcessing() }
+        if (asynchronousProcess) {
+            supervisorScope { async { saveProcessing() } }
         } else {
             saveProcessing()
         }
```
Fabilin, line 332: is there a concrete difference between these two branches?

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -320,22 +389,22 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
...
+    internal suspend fun loadConnectorMessages(ids: List<ConnectorMessageColId>): Map<ConnectorMessageColId, List<ConnectorMessage>> {
         return try {
             if (ids.isEmpty()) {
                 emptyMap()
             } else {
                 connectorMessageCol.find(ConnectorMessageCol::_id `in` ids)
-                    .map { m -> m._id to m.messages.mapNotNull { it?.value as? ConnectorMessage } }
-                    .toMap()
+                    .toList()
+                    .associate { m -> m._id to m.messages.mapNotNull { it?.value as? ConnectorMessage } }
```
Fabilin, line 398: Is there a specific reason for this change?

bot/engine/src/main/kotlin/engine/BotRepository.kt
```diff
@@ -217,24 +218,27 @@ object BotRepository {
...
+            val userTimelineDAO: UserTimelineDAO = injector.provide()
+            val userTimeline = userTimelineDAO.loadWithoutDialogs(botDefinition.namespace, recipientId)
+            val userState = userTimeline.userState
+            val currentState = userState.botDisabled
+
+            if (stateModifier == NotifyBotStateModifier.ACTIVATE_ONLY_FOR_THIS_NOTIFICATION ||
+                stateModifier == NotifyBotStateModifier.REACTIVATE
+            ) {
+                userState.botDisabled = false
+                userTimelineDAO.save(userTimeline, botDefinition)
```
Fabilin, line 231: Not specific to the new code: shouldn't we use the `UserLock` when loading and saving the timeline like this?

bot/engine/src/main/kotlin/engine/user/UserTimelineDAO.kt
Fabilin, line 1: Quite a few breaking changes here, but presumably this interface is not used much outside TOCK

bot/storage-mongo/src/main/kotlin/SendSentenceNotYetLoaded.kt
```diff
@@ -66,7 +67,7 @@ internal class SendSentenceNotYetLoaded(
             if (!messageLoaded) {
                 logger.debug { "load message for $id" }
                 messageLoaded = true
-                loadedMessages.addAll(UserTimelineMongoDAO.loadConnectorMessage(toActionId(), dialogId))
+                loadedMessages.addAll(runBlocking { UserTimelineMongoDAO.loadConnectorMessage(toActionId(), dialogId) } )
```
Fabilin, line 70: Doesn't this code present a risk of race conditions?

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -481,209 +559,232 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
...
-    }
+            dialogCol
+                .distinct(
+                    Stories.actions.state.intent as KProperty1<DialogCol, String>,
+                    and(DialogCol_.ApplicationIds `in` applicationsIds.filter { it.isNotEmpty() })
+                )
+                .toList()
+                //TODO fix kmongo distinct nullable issue
+                .filterNotNull()
+                .toSet()
```
Fabilin:
```suggestion
                .filterNotNullTo(mutableSetOf())
```
very minor suggestion, may not be worth doing if we plan to remove the null check eventually

Fabilin (commented):
Looks good to me in terms of implementation, however this presents quite a few breaking changes. It may be worth adding some backward compatibility bandaids, which we could eventually deprecate and remove.

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
vsct-jburet: yes, in the first branch, the save is processed asynchronously (does not wait the save to return)

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
vsct-jburet: suggested by IDE for simplification

bot/engine/src/main/kotlin/engine/BotRepository.kt
vsct-jburet: No, the user lock is supposed to avoid two concurrent requests (and even in that case, there is a timeout). If you call this internal API directly, you are supposed to know what you are doing

bot/storage-mongo/src/main/kotlin/SendSentenceNotYetLoaded.kt
vsct-jburet: I don't see why

bot/engine/src/main/kotlin/engine/feature/FeatureDAO.kt
vsct-jburet: users are supposed to use BotBus#isFeatureEnabled

bot/engine/src/main/kotlin/engine/user/UserLock.kt
vsct-jburet: UserLock is an internal API. Developers are not supposed to use it directly

bot/storage-mongo/src/main/kotlin/FeatureMongoDAO.kt
vsct-jburet: fixed

bot/engine/src/main/kotlin/engine/user/UserTimelineDAO.kt
vsct-jburet: same perception

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
vsct-jburet: done

bot/engine/src/main/kotlin/engine/feature/FeatureDAO.kt
Fabilin: Ideally yes, however I can attest to a lot of unit tests mocking this method. Code that does not have access to a `BotBus` also needs to go through the DAO directly.

bot/engine/src/main/kotlin/engine/user/UserLock.kt
Fabilin: That's fair

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
Fabilin: I think that's how it was intended at first, but with coroutines, is there a concrete difference between these two implementations?

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
Fabilin: Fair enough, I am not sure this simplifies much but it does not complicate anything either.

bot/storage-mongo/src/main/kotlin/SendSentenceNotYetLoaded.kt
Fabilin: The red flag for me is it sets `messageLoaded` before actually loading the messages. However, given this class is internal and this specific method is not using coroutines yet, I assume it cannot run in parallel.

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
vsct-jburet: yes, because of the async keyword without await

bot/engine/src/main/kotlin/engine/feature/FeatureDAO.kt
vsct-jburet: we release a major version : we can break small things :)

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
Fabilin: Looking better with the new `fireAndForget` utility 👍

PR #1945, opened 2025-11-17
Fix CORS problem on WebConnector (Options method not allowed)

assouktim:
- Apply CorsHandler globally on $path* so that all sub-routes and HTTP methods (including automatic preflight OPTIONS) are covered.
- Remove redundant CorsHandler from child routes (/sse, /sse/direct) since the global handler already applies.


Benvii (approved):

Benvii:
Issue introduced here : https://github.com/theopenconversationkit/tock/commit/6a527624

scezen (approved):

assouktim:
No, we don't use it yet.

PR #1948, opened 2025-11-17
resolves #1947 Add Open AI Connector

### LINKED ISSUE #1947: Add Open AI Connector

vsct-jburet:
@francoisno @Fabilin @Benvii also look at:

- https://github.com/theopenconversationkit/tock-bot-demo/pull/14/files
- https://github.com/theopenconversationkit/tock-docker/pull/54/files


bot/connector-open-ai/src/main/kotlin/OpenAIConnectorProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+                    mandatory = false
+                )
+            )
+        )
+
+    override val supportedResponseConnectorMessageTypes: Set<KClass<out ConnectorMessage>> = setOf(OpenAIConnectorMessage::class)
+}
+
+// used in file META-INF/services/ai.tock.bot.connector.ConnectorProvider
+internal class OpenAIConnectorProviderService : ConnectorProvider by OpenAIConnectorProvider
```
Fabilin: Is there a reason for the dedicated object? Couldn't `OpenAIConnectorProvider` be a regular class so that it can be directly be referenced in the service metadata?

bot/api/client/src/main/kotlin/TockClientBus.kt
```diff
@@ -107,15 +102,11 @@ class TockClientBus(
     override fun defaultDelay(answerIndex: Int): Long = 0
 
     override suspend fun enableStreaming() {
```
Fabilin, line 104: maybe consider throwing an error or returning `false` if streaming cannot be enabled on the current connector?

bot/connector-open-ai/src/main/kotlin/OpenAIConnector.kt
```diff
@@ -0,0 +1,251 @@
...
+val openAIConnectorType = ConnectorType(OPEN_AI_CONNECTOR_ID)
+
+private val corsPattern = property("tock_web_cors_pattern", ".*")
+private val sseEnabled = booleanProperty("tock_web_sse", false)
+
+class OpenAIConnector internal constructor(
+    val connectorId: String,
+    val path: String,
+    private val webSecurityHandler: WebSecurityHandler
+) : ConnectorBase(openAIConnectorType, setOf(CAROUSEL)) {
```
Fabilin, line 73: Does this connector actually support carousels? It does not seem implemented here.

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorCallback.kt
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Locale
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CopyOnWriteArrayList
+
+class OpenAIConnectorCallback(
+    applicationId: String,
+    val locale: Locale,
+    private val context: RoutingContext?,
+    private val actions: MutableList<Action> = CopyOnWriteArrayList(),
+    private val metadata: MutableMap<String, String> = ConcurrentHashMap(),
```
Fabilin, line 44: Is this metadata actually used anywhere?

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+import ai.tock.shared.injector
+import ai.tock.shared.provide
+import ai.tock.shared.resourceAsString
+import ai.tock.shared.security.auth.spi.WebSecurityHandler
+import ai.tock.shared.security.auth.spi.WebSecurityMode
+import kotlin.reflect.KClass
+
+private const val WEB_SECURITY_MODE_PARAM = "web_security_mode"
+
+private val cookieAuth = booleanProperty("tock_web_cookie_auth", false)
```
Fabilin, line 36: Do these make sense for Open WebUI?

bot/connector-open-ai/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>tock-bot</artifactId>
+        <groupId>ai.tock</groupId>
+        <version>25.10.2-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>tock-bot-connector-open-ai</artifactId>
+    <name>Tock Bot Web Connector</name>
```
Fabilin: May need an update

shared/src/main/kotlin/serialization/Json.kt
```diff
@@ -0,0 +1,107 @@
...
+inline fun <reified T : Any> readJson(input: InputStream): T = json.decodeFromStream<T>(input)
+
+inline fun <reified T : Any> readJson(content: String): T = json.decodeFromString(content)
+
+inline fun <reified T> loadResource(path: String): T? =
+    object {}::class.java.getResourceAsStream(path)?.let { stream ->
+        readJson(
+            stream
+                .bufferedReader()
+                .use(BufferedReader::readText),
```
Fabilin: Why not call `readJson(InputStream)` here instead of loading the whole string first?

shared/src/main/kotlin/serialization/Json.kt
Fabilin, line 1: Is there a specific reason to use Kotlin Serialization instead of Jackson? Maintaining two distinct JSON frameworks may become an unnecessary burden if not.

shared/src/main/kotlin/vertx/VertXs.kt
```diff
@@ -230,25 +233,44 @@ internal fun vertxExecutor(): Executor {
...
 }
 
 fun HttpServerResponse.sendSsePing(): CompositeFuture =
     Future.all(
         write("event: ping\n"),
-        write("data: 1\n\n")
+        write("data: 1\n\n"),
+    )
+
+fun HttpServerResponse.sendSseEnd(): CompositeFuture =
```
Fabilin: This seems very specific to OpenAI, shouldn't it go to the dedicated module?

Fabilin (commented):
A couple copy-pasting mistakes I believe, otherwise looks good.

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorProvider.kt
vsct-jburet: I copied this from the web connector, but there is a reason: the class has to be instantiated (see Java Service Locator), and OpenAIConnectorProvider is an object. :)

bot/connector-open-ai/src/main/kotlin/OpenAIConnector.kt
vsct-jburet: fixed

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorCallback.kt
vsct-jburet: yes, for streaming

bot/connector-open-ai/pom.xml
vsct-jburet: fixed

shared/src/main/kotlin/serialization/Json.kt
vsct-jburet: true

bot/api/client/src/main/kotlin/TockClientBus.kt
vsct-jburet: For now, I prefer to ignore it — I don’t want to expose the connector’s behavior to the client. But I agree that I need to document this behavior.

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorProvider.kt
vsct-jburet: I prefer to keep it because the service may be used outside of Open WebUI

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorProvider.kt
Fabilin: Right, that's what I meant: couldn't `OpenAIConnectorProvider` be a regular class instead of an object (so that it can be instantiated by the service locator)? It would avoid the extra class.

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorCallback.kt
Fabilin: Ah yes, missed the new method call 👍

shared/src/main/kotlin/serialization/Json.kt
vsct-jburet: I think kotlinx.serialization is a better target than jackson - because it uses no reflection and it allows multi-platform

shared/src/main/kotlin/vertx/VertXs.kt
vsct-jburet: done

bot/connector-open-ai/src/main/kotlin/OpenAIConnectorProvider.kt
vsct-jburet: done

Fabilin (approved):

PR #1949, opened 2025-11-18
resolves #1946 bot_engine: implement coroutine-based nlp listeners

### LINKED ISSUE #1946: [Bot Engine] allow asynchronous operations in NLP listeners
With an increasing amount of TOCK APIs supporting coroutines, it would make sense to bring support to `NlpListener`s, especially considering its implementations are likely to use network calls and similarly long operations.

Fabilin:
resolves #1946 by adding a new `AsyncNlpListener` interface which is special-cased by its callers. Because suspending and non-suspending methods cannot overload one another, all methods in the new interface had to be renamed.

### DIFF
--- a/bot/engine/src/main/kotlin/engine/nlp/AsyncNlpListener.kt
+++ b/bot/engine/src/main/kotlin/engine/nlp/AsyncNlpListener.kt
@@ -126,7 +126,7 @@ interface AsyncNlpListener {
      * ```
      * object MyNlpListener : AsyncNlpListener {
      *  override suspend fun onError(query: NlpQuery, dialog: Dialog, throwable: Throwable?) {
-     *   super.error(query, dialog, throwable)
+     *   super.onError(query, dialog, throwable)
      *   dialog.state.currentIntent = bot.myCustomErrorStory.mainIntent()
      *  }
      * }


bot/engine/src/main/kotlin/engine/nlp/engine/nlp/AsyncNlpListener.kt
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Used to customize behaviour of NLP parsing - and also to monitor NLP requests - on bot side.
+ * Has to be registered using [ai.tock.bot.engine.BotRepository.registerNlpListener].
+ */
+@ExperimentalTockCoroutines
+interface AsyncNlpListener : NlpListener {
+
+    @Deprecated("Use coroutines to call this interface", replaceWith = ReplaceWith("detectKeyword(sentence)"))
+    override fun handleKeyword(sentence: String): Intent? = runBlocking {
```
vsct-jburet: Why add theses deprecated methods, as this is a new interface ?

bot/engine/src/main/kotlin/engine/nlp/engine/nlp/AsyncNlpListener.kt
```diff
@@ -0,0 +1,208 @@
...
+
+    /**
+     * Used to handle "secret" keywords.
+     *
+     * @return null if no keyword is detected. If not null the nlp call is not started and the returned intent is used.
+     */
+    suspend fun detectKeyword(sentence: String): Intent? = null
+
+    @Deprecated("Use coroutines to call this interface", replaceWith = ReplaceWith("precomputeNlp(sentence, userTimeline, dialog, botDefinition)"))
+    override fun precompute(
```
vsct-jburet: same remark on @Deprecated

bot/engine/src/main/kotlin/engine/nlp/engine/nlp/AsyncNlpListener.kt
Fabilin: It's a subinterface of `NlpListener`, so we need to provide a default implementation

vsct-jburet: Why not doing a fresh new interface and then add the async listener in BotRepository + an internal wrapper to use the old code ?

```kotlin
fun registerNlpListener(listener: AysncNlpListener) {
        registerNlpListener(NlpListenerWrapper(listener))
}
```

Fabilin: Need to do it the other way, because we need to call suspend methods, but true that's a good workaround

PR #1951, opened 2025-11-21
[DERCBOT-1708] Google Chat Intro Message

scezen:
Ticket : [DERCBOT-1727](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1727)

### Details
- Added support for an introductory message automatically sent for the first message of a space/thread.
- Introduced a configurable cooldown (default aligned with dialog validity: 24h) to prevent spam.
- Updated `send()` flow to dispatch the intro message **before** the bot’s actual reply when the cooldown allows it.
- Added a new connector parameter `introMessage` and exposed it in the connector configuration UI.

### DIFF
--- a/bot/connector-google-chat/src/main/kotlin/GoogleChatConnector.kt
+++ b/bot/connector-google-chat/src/main/kotlin/GoogleChatConnector.kt
@@ -78,8 +78,8 @@ class GoogleChatConnector(
                                             spaceName,
                                             threadName,
                                             chatService,
-                                            introMessage
-                                        )
+                                            introMessage,
+                                        ),
                                     ),
                                 )
                             }
@@ -119,8 +119,7 @@ class GoogleChatConnector(
                                 .create(
                                     callback.spaceName,
                                     message.toGoogleMessage().setThread(Thread().setName(callback.threadName)),
-                                )
-                                .setMessageReplyOption("REPLY_MESSAGE_FALLBACK_TO_NEW_THREAD")
+                                ).setMessageReplyOption("REPLY_MESSAGE_FALLBACK_TO_NEW_THREAD")
                                 .execute()
 
                         logger.info { "Google Chat API response: ${response?.name}" }
--- a/bot/connector-google-chat/src/main/kotlin/GoogleChatConnectorCallback.kt
+++ b/bot/connector-google-chat/src/main/kotlin/GoogleChatConnectorCallback.kt
@@ -30,7 +30,6 @@ data class GoogleChatConnectorCallback(
     private val chatService: HangoutsChat,
     private val introMessage: String?,
 ) : ConnectorCallbackBase(applicationId, googleChatConnectorType) {
-
     private val logger = KotlinLogging.logger {}
 
     /**
@@ -49,9 +48,7 @@ data class GoogleChatConnectorCallback(
      * - An intro message is configured
      * - The user timeline has no bot actions (new conversation)
      */
-    private fun shouldSendIntro(userTimeline: UserTimeline): Boolean {
-        return introMessage != null && !userTimeline.containsBotAction()
-    }
+    private fun shouldSendIntro(userTimeline: UserTimeline): Boolean = introMessage != null && !userTimeline.containsBotAction()
 
     /**
      * Sends the intro message to the Google Chat space/thread.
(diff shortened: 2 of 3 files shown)


bot/connector-google-chat/src/main/kotlin/GoogleChatConnector.kt
```diff
@@ -46,13 +47,21 @@ class GoogleChatConnector(
...
+    private val introMessage: String? = null,
 ) : ConnectorBase(GoogleChatConnectorProvider.connectorType) {
     private val logger = KotlinLogging.logger {}
     private val executor: Executor by injector.instance()
 
+    // Align intro cooldown with dialog validity (24 hours)
+    private val introCooldownSeconds: Long = longProperty("tock_bot_dialog_max_validity_in_seconds", 60 * 60 * 24)
+
+    // Last intro send time per space + thread
+    private val introSentThreads: MutableMap<String, Long> = mutableMapOf()
```
Benvii: High risk of memory leak here, the state stored here will not be shared across all chat-api nodes.

As discussed yesterday we need to communicate data from the ConnectorController to the Connector. Why ? Because the [ConnectorController load the UserTimeline](https://github.com/CreditMutuelArkea/tock/blob/def2cab2b41ac974c05fce6ff76fef4a22bcf2c7/bot/engine/src/main/kotlin/engine/TockConnectorController.kt#L181) and we can rely on it as it contains a list of dialogs to have the associated Google Chat thread conversation history.

Currently we have ConnectorData to send data from the connector to the ConnectorController, when the ConnectorController needs to communicate with the Connector it uses each connector callback implementation of ConnectorCallbackBase, in our case GoogleChatConnectorCallback. Callback is used to communicated different stuff during the Action lifecycle treatment, in our case we want to tell the GoogleChatConnectorCallback connector when the UseTimeline is loaded with it so that it can extract if it's en empty / first dialog or not and the send the intro message.

Add to ConnectorCallbackBase a new method called something like initialUserTimelineLoaded, call it from the controller. In GoogleChatConnectorCallback send the intro message in this method if the dialog is empty.

bot/connector-google-chat/src/main/kotlin/builder/GoogleChatCardMessageBuilders.kt
```diff
@@ -46,7 +46,9 @@ internal const val GOOGLE_CHAT_ACTION_INTENT_PARAMETER = "INTENT"
 annotation class CardElementMarker
 
 @CardElementMarker
-abstract class ChatCardElement(val i18nTranslator: I18nTranslator) : I18nTranslator by i18nTranslator {
+abstract class ChatCardElement(
+    val i18nTranslator: I18nTranslator,
+) : I18nTranslator by i18nTranslator {
```
Benvii, line 51: Is it the linter that is generating all those changes ?

bot/connector-google-chat/src/main/kotlin/GoogleChatConnectorProvider.kt
```diff
@@ -75,16 +92,58 @@ internal object GoogleChatConnectorProvider : ConnectorProvider {
...
+        val sourceCredentials = getSourceCredentials(connectorConfiguration)
+        return ImpersonatedCredentials.create(
+            sourceCredentials,
+            targetServiceAccount,
+            null,
+            listOf(CHAT_SCOPE),
+            3600,
+            null,
+        )
+    }
```
Benvii, line 132: I have red that in the other pull request, rebase it plz.

bot/connector-google-chat/src/main/kotlin/builder/GoogleChatCardMessageBuilders.kt
scezen: Yes... I had to format all files with KTLint for the CI to pass

bot/connector-google-chat/src/main/kotlin/GoogleChatConnectorCallback.kt
```diff
@@ -18,9 +18,66 @@ package ai.tock.bot.connector.googlechat
...
+                chatService
+                    .spaces()
+                    .messages()
+                    .create(
+                        spaceName,
+                        GoogleChatConnectorTextMessageOut(introMessage)
+                            .toGoogleMessage()
+                            .setThread(Thread().setName(threadName)),
+                    ).setMessageReplyOption("REPLY_MESSAGE_FALLBACK_TO_NEW_THREAD")
+                    .execute()
```
Benvii, line 74: Did you check if the message is correctly logged into the user's dialog / timeline ? and then visible into the studio dialogs ?

PR #1954, opened 2025-12-02
[DERCBOT-1746] Langfuse V3 Callback Fixes

scezen:
Ticket : [DERCBOT-1746](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1746)

### Details
- Migrated from deprecated `langfuse.callback` API to the new `langfuse.langchain` module.
- Updated the `LangfuseCallbackHandlerFactory` to align with the new Langfuse V3 client.
- Added internal caching for the Langfuse client to avoid repeated initializations.
- Updated the observability service to use the new `get_client()` API and revised trace extraction logic.
- Adjusted the RAG chain execution to pass metadata (user, session, tags) using the new supported mechanism.
- Cleaned up unused parameters and removed unsupported kwargs that Langfuse no longer handles.


gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/factories/callback_handlers/langfuse_callback_handler_factory.py
```diff
@@ -59,30 +60,81 @@ class LangfuseCallbackHandlerFactory(LangChainCallbackHandlerFactory):
...
+        self._get_langfuse_client()
+
+        # Ignore Langfuse V2 parameters to stay backward-compatible
+        if kwargs:
+            logger.debug(
+                'Ignoring unsupported Langfuse CallbackHandler kwargs in V3: %s',
+                list(kwargs.keys()),
+            )
+
+        return LangfuseCallbackHandler(
```
Benvii, line 95:
```suggestion
        # Langfuse SDK maintains an internal map / pool of clients based on there public key, that why the client isn't passed to the callbackhandler constructor.
        return LangfuseCallbackHandler(
```

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/observability/observabilty_service.py
```diff
@@ -45,13 +46,20 @@ def check_observability_setting(setting: ObservabilitySetting) -> bool:
...
-        return None
\ No newline at end of file
+    if not isinstance(observability_handler, LangfuseCallbackHandler):
+        return None
+
+    trace_id = getattr(observability_handler, 'last_trace_id', None)
+    if trace_id is None:
+        return None
+
+    langfuse_client = get_client()
```
Benvii: Are you sure you are getting the right client here ?
You should be using the client public key to be sure at least.
You can access the client using observability_handler.client I think it's the easiest way.


This could lead to rights access errors.

assouktim (approved):

PR #1969, opened 2025-12-22
[DERCBOT-1676] Fix dialog date filters and add comprehensive tests

antoine-billiet-arkea-partnre:
This pull request introduces new support for filtering dialogs by their activity period overlap, in addition to creation date, and provides a reusable DSL for building MongoDB aggregation expressions. It also adds comprehensive unit tests to ensure correct mapping of date parameters in dialog search queries.

**Dialog filtering enhancements:**

* Added `dialogActivityFrom` and `dialogActivityTo` fields to both `DialogsSearchQuery` and `DialogReportQuery` to allow filtering dialogs based on whether their activity period (from first to last action) overlaps a given range. This is documented and mapped in the conversion method. [[1]](diffhunk://#diff-71d501569496c173db5bb542d8c1945ebbfefcd4333853191dab74b438a3f14fR53-R59) [[2]](diffhunk://#diff-71d501569496c173db5bb542d8c1945ebbfefcd4333853191dab74b438a3f14fR89-R90) [[3]](diffhunk://#diff-2b2e5b2d5ee5e6c492584a74118988ad8dcdebdfc72ebcf947aebed5b5d40255R72-R78)

**MongoDB aggregation improvements:**

* Introduced a new `MongoAgg` object providing a DSL for building MongoDB aggregation expressions, including helpers for date aggregation in nested arrays and for filtering by creation date or activity period overlap. This makes aggregation queries more readable and maintainable.

**Testing:**

* Added a new test class `DialogSearchServiceTest` to verify that all date filtering parameters (creation and activity periods) are correctly mapped from `DialogsSearchQuery` to `DialogReportQuery`, including handling of nulls and combinations.

**Code organization:**

* Updated imports in `UserTimelineMongoDAO.kt` to include explicit Kotlin standard library imports for better clarity and maintainability.


bot/storage-mongo/src/main/kotlin/MongoAgg.kt
```diff
@@ -0,0 +1,360 @@
...
+ *     initialValue = null,
+ *     `in` = MongoAgg.cond(
+ *         MongoAgg.eq(MongoAgg.value(), null),
+ *         MongoAgg.ifNull(MongoAgg.min("${MongoAgg.thisVar()}.actions.date"), null),
+ *         MongoAgg.min(listOf(MongoAgg.value(), MongoAgg.ifNull(MongoAgg.min("${MongoAgg.thisVar()}.actions.date"), MongoAgg.value())))
+ *     )
+ * ))
+ * ```
+ */
+object MongoAgg {
```
assouktim: This utility class is not necessary in this case.
The functional requirement is simply to answer the following question:
“Is there at least one action in all the dialogue stories whose date meets the condition?”

The current code already meets this requirement in a simple, readable, and efficient way, without introducing additional logic to process arrays via $reduce, $cond, or $ifNull.

Stories.actions.date lt dialogCreationDateTo?.toInstant(): This filter, for example, applies directly to the field contained in the stories and actions arrays.
MongoDB will therefore match documents that contain at least one action whose date meets the condition.

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -761,16 +807,14 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
...
-                                        null
-                                    } else {
-                                        Stories.actions.date gt dialogCreationDateFrom?.toInstant()
-                                    },
-                                    if (dialogCreationDateTo == null) {
-                                        null
-                                    } else {
-                                        Stories.actions.date lt dialogCreationDateTo?.toInstant()
-                                    },
+                                    buildDialogCreationDateFilter(
```
assouktim: The gte and lte operators must be used to avoid the strict nature of the condition, especially since filtering is now performed at specific times.

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -130,7 +130,53 @@ import java.time.Instant.now
...
+import kotlin.collections.toSet
+import kotlin.let
 import kotlin.reflect.KProperty1
+import kotlin.takeIf
+import kotlin.text.isNotEmpty
+import kotlin.text.isNullOrBlank
+import kotlin.text.take
+import kotlin.text.trim
+import kotlin.to
+import kotlin.with
```
assouktim: We discussed this block:
if (from == null) null else DialogCol_.LastUpdateDate gt from?.toInstant(), if (to == null) null else DialogCol_.LastUpdateDate lt to?.toInstant(),

we said that it was redundant and, above all, useless because the from and to contain dead code. Please double-check, and if that's correct, remove them from the query and filtering.

bot/storage-mongo/src/test/kotlin/UserTimelineMongoDAOTest.kt
```diff
@@ -94,4 +206,345 @@ internal class UserTimelineMongoDAOTest : AbstractTest() {
                 UserTimelineMongoDAO.loadWithoutDialogs("namespace", newId).playerId,
             )
         }
+
```
assouktim, line 111: I will validate this test later on. I am waiting for your feedback on my first review, as it will significantly impact this unit test.

bot/admin/server/src/main/kotlin/model/DialogsSearchQuery.kt
```diff
@@ -50,6 +50,13 @@ data class DialogsSearchQuery(
     val annotationCreationDateTo: ZonedDateTime? = null,
     val dialogCreationDateFrom: ZonedDateTime? = null,
     val dialogCreationDateTo: ZonedDateTime? = null,
+    /**
+     * Filter dialogs by activity period overlap.
+     * A dialog is included if its activity period (from first to last action) overlaps the filter range.
```
assouktim: I'm not sure that's right. Normally, at least one action must be within the range, not all of the actions in the dialogue.

A dialogue must be included if at least one action is within the range.

bot/storage-mongo/src/main/kotlin/MongoAgg.kt
antoine-billiet-arkea-partnre: ---

This PR addresses **two distinct filtering requirements**:

---

### 1. Filter by Dialog Creation Date (`dialogCreationDateFrom` / `dialogCreationDateTo`)

**Requirement:** Filter dialogs where the **creation date** (= the date of the first/oldest action) falls within a specified period.

**Condition:** `creationDateFrom <= min(actions.date) <= creationDateTo`

For the upper bound (`dialogCreationDateTo`), the simple approach works because if `min(date) <= toDate`, then at least one action exists with `date <= toDate`.

However, **for the lower bound (`dialogCreationDateFrom`), the simple approach fails:**

**Counter-example:**
- Dialog with actions at dates: `[2025-01-01, 2025-01-15, 2025-01-30]`
- Filter: `dialogCreationDateFrom = 2025-01-10` (dialogs created from January 10th)

| Approach | Result | Correct? |
|----------|--------|----------|
| Simple (`Stories.actions.date gte 2025-01-10`) | Matches (Jan 15th action satisfies) | Wrong |
| Aggregation (`min(actions.date) >= 2025-01-10`) | No match (Jan 1st < Jan 10th) | Correct |

The dialog was **created on January 1st**, so it should NOT be returned when filtering for dialogs created from January 10th.

**The `Agg` class with `$reduce`/`$min` is required for this filter.**

---

### 2. Filter by Dialog Activity (`dialogActivityFrom` / `dialogActivityTo`)

**Requirement:** Filter dialogs that had **activity** during the specified period, meaning:
- At least one action **on or after** the lower bound (`activityFrom`)
- At least one action **before** the upper bound (`activityTo`)

These two conditions **do not need to be on the same action**.

**Condition:**
- `∃ action >= activityFrom` (at least one action after the lower bound)
- `∃ action < activityTo` (at least one action before the upper bound)

**Mathematical equivalence:**

| Statement | Equivalent to |
|-----------|---------------|
| `∃ action >= activityFrom` | `max(actions.date) >= activityFrom` |
| `∃ action < activityTo` | `min(actions.date) < activityTo` |

This means **both approaches produce the same result** for the activity filter:

```kotlin
// Simple approach
Stories.actions.date gte activityFrom
Stories.actions.date lt activityTo

// Aggregation approach (current implementation)
Agg.filterByPeriodOverlap(...)  // max >= from AND min < to
```

You are correct that the simple approach would work for this filter. I chose the aggregation approach for **consistency** with the creation date filter and to make the intent explicit (computing min/max), but both are functionally equivalent.

---

### Summary

| Filter | Requirement | Simple approach works? | `Agg` required? |
|--------|-------------|------------------------|-----------------|
| **dialogCreationDate** | `min(actions.date)` within period | No (fails for `from`) | **Yes** |
| **dialogActivity** | ∃ action >= from AND ∃ action < to | **Yes** | No (but used for consistency) |

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
antoine-billiet-arkea-partnre: You're right, there is an inconsistency in the date filtering operators across the codebase.

### Current state of date filters

| Filter | `from` operator | `to` operator | Interval |
|--------|-----------------|---------------|----------|
| **annotationCreationDate** | `gt` (exclusive) | `lt` (exclusive) | `(from, to)` |
| **dialogCreationDate** | `gte` (inclusive) | `lte` (inclusive) | `[from, to]` |
| **dialogActivity** | `gte` (inclusive) | `lt` (exclusive) | `[from, to)` |

### The problem

Three different conventions for three different filters:
- **Annotations**: Open interval `(from, to)` — both bounds are **exclusive**
- **Creation date**: Closed interval `[from, to]` — both bounds are **inclusive**
- **Activity**: Half-open interval `[from, to)` — lower bound inclusive, upper bound exclusive

This is indeed confusing and error-prone, especially when filtering at specific timestamps.

### Proposed harmonization

I suggest adopting **`[from, to]`** (closed interval with `gte`/`lte`) for all date filters:
- More intuitive for users: "from January 1st to January 31st" includes both dates
- Consistent behavior across all filters
- Avoids edge cases where exact timestamp matches are unexpectedly excluded

I will update:
1. `annotationCreationDate`: change `gt` → `gte` and `lt` → `lte`
2. `dialogActivity`: change `lt` → `lte` (and update `MongoAgg.filterByPeriodOverlap`)

Do you agree with this approach, or would you prefer a different convention (e.g., half-open `[from, to)` everywhere)?

bot/admin/server/src/main/kotlin/model/DialogsSearchQuery.kt
antoine-billiet-arkea-partnre: see first comment.

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
assouktim: Yes, a closed interval with gte/lte for all date filters.

bot/storage-mongo/src/main/kotlin/MongoAgg.kt
assouktim: I prefer the simplest option.

PR #1975, opened 2025-12-30
[DERCBOT-1404] Fix locale string interpretation

scezen:
Ticket : [DERCBOT-1404](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1404)

### Details
- Fixed locale parsing issue in `RestConnector` where legacy formats (ex: `fr_FR`) resulted in empty locale data.
- Input locales are now sanitized (replacing `_` with `-`) to ensure compatibility with BCP 47 standards before parsing.
- `nlpStats` in Studio Analytics now correctly reflect the provided locale when testing via TVK or Postman.


bot/connector-rest/src/main/kotlin/RestConnector.kt
```diff
@@ -205,4 +205,22 @@ class RestConnector(
                 else -> getTargetConnector(targetConnectorType)?.toConnectorMessage(message)?.invoke(this) ?: emptyList()
             }
         }
-}
+
+    private fun parseLocale(rawLocale: String): Locale {
```
assouktim: I think it's better to fix the problem at its source, from the caller.
Here, the code attempts a Locale.forLanguageTag(). The caller must pass this tag instead of what is currently being done, which is causing the problem, a Retrofit toString.

Look here: ConnectorRestClient#talk.

You will need to change the parameter from Local to string, then make the call as follows: 

`return getService(path).talk(locale.toLanguageTag(), query).execute()`

assouktim: After that, you should notice that the language display in the user interface is not correct. To fix this, use the normalizeLocaleCode function.

assouktim (changes requested):

assouktim (changes requested):
.

PR #1983, opened 2026-01-09
fixe typing indicator whatsapp cloud

SNCFamiraBenkaci:
Sends a "Typing Indicator" event to the WhatsApp Cloud API.
This request is used to notify the end user that the business
is currently typing a message ("typing…") in a WhatsApp conversation.
The typing indicator is linked to an existing message via its `message_id`
and does not send a visible message to the user.


vsct-jburet (approved):

bot/connector-whatsapp-cloud/src/main/kotlin/WhatsAppCloudApiClient.kt
asus007, line 1: Eviter de changer de version juste pour le typing indicator (car cela peut impacter tous les autres appels graph api)

SNCFamiraBenkaci: ok merci, je vais changer ça

PR #2000, opened 2026-02-06
fix #1998 [Bot Engine] deletes builtin not founds

### LINKED ISSUE #1998: [engine] the otherStarterIntents is not working
With @Fabilin, we noticed that the otherStarterIntents property is not being consumed by the engine.

As a result, the story is not being triggered.

TiagoCerqueira:


bot/engine/src/main/kotlin/engine/config/BotDefinitionWrapper.kt
```diff
@@ -103,6 +109,10 @@ internal class BotDefinitionWrapper(val botDefinition: BotDefinition) : BotDefin
                 .values.flatten()
 
         this.allStoriesById = allStories.associateBy { it.id }
+        val builtinStoriesNotFounds = botDefinition.stories.map(StoryDefinition::id) - configuredStories.map(StoryDefinitionConfiguration::storyId).toSet()
```
Fabilin: I believe this should rather be something like `configuredStories.filter { it.currentType == builtin && it.storyId != definition.unknownStory.storyId && definition.builtInStory(it.storyId) == definition.unknownStory }`

Thoughts @zigzago?

zigzago: agreed

PR #2001, opened 2026-02-10
[DERCBOT-1783] Admin: add evaluation sample management API

antoine-billiet-arkea-partnre:
- Add EvaluationSample and Evaluation models with all required fields
- Add ActionRef, EvaluationReason, EvaluationsResult, Evaluator models
- Implement EvaluationSampleDAO and EvaluationDAO interfaces
- Implement MongoDB DAOs with proper indexing
- Add EvaluationService with sample creation, evaluation, and status management
- Add EvaluationVerticle with REST endpoints for CRUD operations
- Add unit tests for service and integration tests for DAOs
- Support pagination for action-refs with status filtering
- Add evaluationsResult computation (total, evaluated, remaining, positiveCount, negativeCount)

### DIFF
--- a/bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
+++ b/bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
@@ -86,11 +86,9 @@ import com.mongodb.client.model.Accumulators.sum
 import com.mongodb.client.model.Aggregates.sample
 import com.mongodb.client.model.Filters.elemMatch
 import com.mongodb.client.model.Filters.eq
-import com.mongodb.client.model.Filters.exists
 import com.mongodb.client.model.Filters.gte
 import com.mongodb.client.model.Filters.`in`
 import com.mongodb.client.model.Filters.lte
-import com.mongodb.client.model.Filters.ne
 import com.mongodb.client.model.IndexOptions
 import kotlinx.coroutines.runBlocking
 import mu.KotlinLogging
@@ -1276,7 +1274,7 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
                         filtersAnd(
                             `in`("type", listOf("sentence", "sentenceWithFootnotes")),
                             filtersRegex("text", "\\S"),
-                        )
+                        ),
                     ),
                 ),
             ),


bot/storage-mongo/target/generated-sources/kapt/compile/ai/tock/bot/admin/bot/BotApplicationConfiguration_.kt
vsct-jburet, line 1: The generated files are intentionally committed in order to track changes made by the generators

vsct-jburet: not blocking but if tou have 2s to commit them again ;)

bot/admin/server/src/main/kotlin/verticle/EvaluationVerticle.kt
```diff
@@ -0,0 +1,99 @@
...
+        private const val PATH_SAMPLE = "/bots/:botId/evaluation-samples/:sampleId"
+        private const val PATH_ACTION_REFS = "/bots/:botId/evaluation-samples/:sampleId/action-refs"
+        private const val PATH_EVALUATION = "/bots/:botId/evaluation-samples/:sampleId/evaluations/:evaluationId"
+        private const val PATH_CHANGE_STATUS = "/bots/:botId/evaluation-samples/:sampleId/change-status"
+    }
+
+    fun configure(verticle: WebVerticle) {
+        with(verticle) {
+            // GET /bots/:botId/evaluation-samples - List all evaluation samples for a bot
+            blockingJsonGet(PATH_SAMPLES, botUser) { context ->
```
assouktim, line 43: For all endpoints, use `checkNamespaceAndExecute` to check the request scope.

bot/admin/server/src/main/kotlin/service/EvaluationService.kt
```diff
@@ -0,0 +1,307 @@
...
+        val dialogQuery =
+            DialogReportQuery(
+                namespace = namespace,
+                nlpModel = botId,
+                dialogActivityFrom = request.dialogActivityFrom.atZone(ZoneOffset.UTC),
+                dialogActivityTo = request.dialogActivityTo.atZone(ZoneOffset.UTC),
+                displayTests = request.allowTestDialogs,
+                withAnnotations = false,
+                start = 0L,
+                size = request.requestedDialogCount * 10,
```
assouktim: Is there a reason why we are searching 10 times more dialogues than requested?
I think the size can simply be set to the maximum.

bot/admin/server/src/main/kotlin/service/EvaluationService.kt
```diff
@@ -0,0 +1,307 @@
...
+        val dialogResult = dialogReportDAO.search(dialogQuery)
+        val totalDialogCount = dialogResult.total.toInt()
+
+        val actionRefs = mutableListOf<ActionRef>()
+        val selectedDialogs = mutableSetOf<String>()
+
+        for (dialog in dialogResult.dialogs.shuffled()) {
+            if (selectedDialogs.size >= request.requestedDialogCount) break
+
+            val botActions = dialog.actions.filter { it.playerId.type == PlayerType.bot }
```
assouktim: There are bot-type actions, but they are “debug” type. They should not be reported.

bot/admin/server/src/main/kotlin/service/EvaluationService.kt
```diff
@@ -0,0 +1,307 @@
...
+                size = request.requestedDialogCount * 10,
+            )
+
+        val dialogResult = dialogReportDAO.search(dialogQuery)
+        val totalDialogCount = dialogResult.total.toInt()
+
+        val actionRefs = mutableListOf<ActionRef>()
+        val selectedDialogs = mutableSetOf<String>()
+
+        for (dialog in dialogResult.dialogs.shuffled()) {
```
assouktim: It would be much more efficient to do the randomization on the database side, rather than in memory :

- You retrieve requestedDialogCount * 10.
- You mix them in memory.
- You take the first X valid ones.
- This is not optimal, the database can do this itself.

I suggest updating the dialog search function, adding a random parameter query, and doing the following: 


```kotlin
val list = if (query.random) {
    // Random mode: $sample in the aggregation pipeline
    c.aggregate(
        listOf(
            match(filter),
            sample(size.toInt())
        )
    )
} else {
    // Paginated/sorted mode
    c.find(filter)
        .sort(sortBson) 
        .skip(start.toInt())
        .limit(size)
}
    .toList()
    .map { it.toDialogReport() }

// Construction of the result
val result = if (query.random) {
    DialogReportQueryResult(
        total = count,  
        start = 0L,
        end = 0L,
        dialogs = list
    )
} else {
    DialogReportQueryResult(
        total = count,
        start = start,
        end = start + list.size,
        dialogs = list
    )
}
```

bot/admin/server/src/main/kotlin/service/EvaluationService.kt
```diff
@@ -0,0 +1,307 @@
...
+            if (selectedDialogs.size >= request.requestedDialogCount) break
+
+            val botActions = dialog.actions.filter { it.playerId.type == PlayerType.bot }
+            if (botActions.isNotEmpty()) {
+                selectedDialogs.add(dialog.id.toString())
```
assouktim: Once you have retrieved the dialog from the database, you can apply this code directly, which avoids loops.

```kotlin
val actionRefs = selected.flatMap { dialog ->
    dialog.actions
        .filter { it.playerId.type == PlayerType.bot } // TODO : Add a filter to debug actions
        .map { action ->
            ActionRef(
                dialogId = dialog.id.toString(),
                actionId = action.id.toString()
            )
        }
}
```

bot/admin/server/src/main/kotlin/service/EvaluationService.kt
```diff
@@ -0,0 +1,307 @@
...
+            EvaluationSample(
+                botId = botId,
+                namespace = namespace,
+                name = request.name,
+                description = request.description,
+                dialogActivityFrom = request.dialogActivityFrom,
+                dialogActivityTo = request.dialogActivityTo,
+                requestedDialogCount = request.requestedDialogCount,
+                dialogsCount = selectedDialogs.size,
+                totalDialogCount = totalDialogCount,
```
assouktim, line 126: What happens if the total number of dialog action is zero? 
I don't think it makes sense to create a sample.

bot/storage-mongo/src/main/kotlin/EvaluationMongoDAO.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.litote.kmongo.group
+import org.litote.kmongo.limit
+import org.litote.kmongo.match
+import org.litote.kmongo.save
+import org.litote.kmongo.skip
+import org.litote.kmongo.sum
+
+private val logger = KotlinLogging.logger {}
+
+internal object EvaluationMongoDAO : EvaluationDAO {
```
assouktim, line 48: When we delete a bot, we delete all associated data. See `BotAdminService#deleteApplication()`

bot/admin/server/src/main/kotlin/service/EvaluationService.kt
antoine-billiet-arkea-partnre: see above ligne 114-119 . Frontend ask for a specific code for this particular case

bot/storage-mongo/src/main/kotlin/DialogCol.kt
```diff
@@ -286,6 +292,16 @@ internal data class DialogCol(
         }
 
         abstract fun toAction(dialogId: Id<Dialog>): Action
+
+        fun isEvaluable(): Boolean {
+            if (playerId.type != PlayerType.bot) return false
+            return when (this) {
+                is SendDebugMongoWrapper -> false
+                is SendSentenceMongoWrapper -> !text.isNullOrBlank()
```
assouktim, line 300: Same logic, same code ;)
As with `SendSentenceWithFootnotesMongoWrapper`, you can use `text.isNotBlank()`

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -1224,6 +1256,40 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
...
+            "stories",
+            elemMatch(
+                "actions",
+                filtersAnd(
+                    eq("playerId.type", "bot"),
+                    filtersOr(
+                        eq("type", "choice"),
+                        eq("type", "attachment"),
+                        eq("type", "location"),
+                        filtersAnd(
```
assouktim, line 1274: I see a lot of redundant conditions, especially when checking the text (exists, ne(null), ne(“”), regex). We can simplify this by using a single regex condition to detect non-empty/non-blank text

`filtersRegex(“text”, “\\S”)` // is sufficient

bot/storage-mongo/src/main/kotlin/DialogCol.kt
antoine-billiet-arkea-partnre: !text.isNullOrBlank() in sentence because text is "String?"
text.isNotBlanck() in SendSentenceWithFootnotesMongoWrapper, text type is "String"

so, the alternative with the text.isNotBlanck() for SendSentenceMongoWrapper would be : 

` is SendSentenceMongoWrapper -> text?.isNotBlank() == true`

assouktim: Okay, I see. In that case, you can leave the code as it is. Thank you.

PR #2005, opened 2026-02-16
[DERCBOT-1791] Expose full footnote metadata and title fallback flag

scezen:
Ticket : [DERCBOT-1791](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1791)

### Goal 
Let the front-end reliably hide footnotes with fallback titles without relying on hardcoded strings, while still exposing full document metadata for future filtering needs.

### Details
- Added isTitleFallback to RAG footnotes and propagated it from the orchestrator to the bot engine and web connector (camelCase on the client side via snake_case mapping).
- Exposed full metadata on footnotes so the front-end can filter using any ingested field (including is_title_fallback).
- Aligned the server-side mapping to read is_title_fallback from document metadata and include it in the footnote payload.

Note :
- We keep the existing response structure and only add optional fields (backward compatible).
- metadata keeps original keys from ingestion (often snake_case), while isTitleFallback is a typed convenience field.

### DIFF
--- a/bot/connector-mattermost/src/main/kotlin/MattermostMessageConverter.kt
+++ b/bot/connector-mattermost/src/main/kotlin/MattermostMessageConverter.kt
@@ -64,6 +64,7 @@ internal object MattermostMessageConverter {
                                 footnote.url,
                                 footnote.content,
                                 footnote.score,
+                                footnote.metadata,
                             )
                         },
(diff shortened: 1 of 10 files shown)


assouktim (approved):
Thanks!

PR #2006, opened 2026-02-23
[Gen AI Orchestrator] Fix Python library versions and TU correction

assouktim:


gen-ai/orchestrator-server/src/main/python/server/tests/routers/test_llm_providers_router.py
```diff
@@ -28,41 +24,32 @@
...
 def test_get_llm_provider_by_id():
     """Test getting provider by id (use first provider listed provider class)."""
-    response = client.get(f'{urls_prefix}/{list(LLMProvider)[0]}')
+    provider_id = list(LLMProvider)[0].value
+    response = client.get(f"{urls_prefix}/{provider_id}")
     assert response.status_code == 200
-    assert response.json()['provider'] == list(LLMProvider)[0]
-
-
-def test_get_llm_provider_by_id_wrong_id():
```
antoine-billiet-arkea-partnre, line 43: Please add the unit test
```
def test_get_llm_provider_by_id_wrong_id():
    """Test getting provider by id, with an id that does not exist."""
    response = client.get(f"{urls_prefix}/wrong_id")
    assert response.status_code == 400
```

assouktim: Oh yes, I deleted it by mistake, thank you.

PR #2024, opened 2026-03-30
[DERCBOT-1839] Fix broken Playground

scezen:
Ticket: [DERCBOT-1839](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1839)

  ### Details

  - Fixed Playground error when Langfuse is enabled.
  - The orchestrator was building the observability response without a trace name.
  - We now set the Playground trace name explicitly before returning the response.


assouktim (approved):
I reviewed it and applied the necessary fixes. Thanks!

rkuffer (approved):

PR #2028, opened 2026-04-09
[DERCBOT-1859] Allow label for namespaces

scezen:
Ticket : [DERCBOT-1859](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1859)

### Goal

Add a label for namespaces, with a fallback to the technical namespace when no label is defined.

### Details

- Added an optional `label` field in `NamespaceConfiguration`
- `NamespaceConfiguration` (was used for Sharing Settings option and created in Mongo only when one option was activated in namespace settings) is now created automatically when a namespace is created
- The backend now returns a resolved label for namespaces:
  - Custom label if defined
  - Otherwise the technical namespace
- Updated the main namespace displays in the Studio to show the label when present

### Note
- No breaking change: if no label is set, the UI still works with the technical namespace


assouktim (changes requested):
Thanks! Looks good to me.
One thing to improve though: updating a label requires a browser refresh for the change to be visible at the namespace list and in the header.

rkuffer (approved):
Ok for me for the front part of this PR

assouktim (approved):
Thanks!

PR #2030, opened 2026-04-13
[DERCBOT-1865] Make the compressor fault-tolerant

assouktim:
The Compressor has been made more robust and non-blocking. All errors (API, parsing, labels) are now logged without throwing exceptions, ensuring that the RAG pipeline always continues to run. If the reranking service fails, an automatic fallback returns the input documents.

The selection logic has been simplified and improved with a `fill_to_max_documents` mode. Documents are first filtered based on `min_score`, then supplemented with the best remaining documents if necessary to reach `max_documents`. This approach ensures both compliance with the quality threshold and the completeness of the results.

### DIFF
--- a/gen-ai/orchestrator-core/src/main/kotlin/ai/tock/genai/orchestratorcore/models/compressor/BloomzDocumentCompressorSetting.kt
+++ b/gen-ai/orchestrator-core/src/main/kotlin/ai/tock/genai/orchestratorcore/models/compressor/BloomzDocumentCompressorSetting.kt
@@ -19,10 +19,12 @@ package ai.tock.genai.orchestratorcore.models.compressor
 data class BloomzDocumentCompressorSetting(
     override val maxDocuments: Int,
     override val minScore: Float,
+    override val fillToMaxDocuments: Boolean = false,
     val label: String,
     val endpoint: String,
 ) : DocumentCompressorSettingBase(
         provider = DocumentCompressorProvider.BloomzRerank,
         maxDocuments = maxDocuments,
         minScore = minScore,
+        fillToMaxDocuments = fillToMaxDocuments,
     )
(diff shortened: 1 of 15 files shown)


gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/impls/document_compressor/bloomz_rerank.py
```diff
@@ -74,52 +69,68 @@ def compress_documents(
...
-                        error=str(response.status_code),
-                        cause=f"Response: {response.text}, Reason: {response.reason}",
-                        request=f"[POST] {url}",
-                    )
-                )
-        except GenAIDocumentCompressorErrorException:
-            # Re-raise GenAIDocumentCompressorErrorException without modification
-            raise
+                logger.warning('[Compressor] Fallback to original documents')
+                return documents
```
scezen, line 98: Could an invalid compressor config now pass the check?                                                                                                                                                   
We fallback to the original documents here instead of raising an error, and check_document_compressor_setting() seems to return True after calling compress_documents()

gen-ai/orchestrator-core/src/main/kotlin/ai/tock/genai/orchestratorcore/models/compressor/BloomzDocumentCompressorSetting.kt
```diff
@@ -19,10 +19,12 @@ package ai.tock.genai.orchestratorcore.models.compressor
 data class BloomzDocumentCompressorSetting(
     override val maxDocuments: Int,
     override val minScore: Float,
+    override val fillToMaxDocuments: Boolean = false,
```
scezen, line 22: I think fillToMaxDocuments is true in the UI and also true in the Python model, but it is false here.                                                                                                                                    
Could this create different behavior, maybe for older stored configurations?

rkuffer (approved):
Sounds good to me for the front section.
I ran a few tests, and it looks promising.

gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/impls/document_compressor/bloomz_rerank.py
assouktim: Yes, exactly. Thank you for your very insightful comment.

scezen (approved):

PR #2042, opened 2026-04-21
[Tock Studio] Improvements to dataset creation

rkuffer:
## Summary
Improve form validation feedback in the dataset create/edit dialog.

## Changes
- Auto-scroll to the first invalid field on failed submission (name, description, questions)
- Add `danger` status and error message on ground truth textareas when they exceed the character limit
- Fix `invalidQuestionCount` getter to account for invalid ground truth fields, not just question fields

## Why
Previously, validation errors were silent for fields outside the visible scroll area, leaving the user with no indication of what was blocking submission. Ground truth fields also had no visual feedback on error.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "js/ts.tsdk.path": "bot/admin/web/node_modules/typescript/lib"
+}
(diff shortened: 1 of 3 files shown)


scezen (approved):

assouktim (approved):
Thanks!

PR #2045, opened 2026-04-27
[DERCBOT-1876] Fix dataset RAG technical failures and Mongo snapshot append

scezen:
Ticket : [DERCBOT-1876](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1876)

### Goal
Fix dataset runs that were marked as successful even when the RAG call returned a technical error.

### Details
- Detect RAG technical errors from debug messages returned by Bot API
- Retry the dataset question when this technical error is detected, instead of marking it as completed because a userActionId exists
- Store the final error on failed dataset questions and expose it in the actions endpoint
- Display the backend error message in the dataset run UI
- Append Mongo dialog snapshots atomically to avoid overwriting existing snapshots


assouktim (approved):
Thanks !

rkuffer (approved):

PR #2046, opened 2026-04-29
[DERCBOT-1822] Fix 'number of messages' chart in custom metrics

scezen:
Ticket : [DERCBOT-1822](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1822)

### Goal
Fix the “Number of user messages” chart in Custom Metrics so it counts real user messages, instead of relying on flow transition stats that can overcount messages.

### Details
- Count user messages from dialog user actions, which is the same source of truth used by dialog stats
- Add a dedicated endpoint for user messages by date: `/dialogs/stats/messages-by-date`
- Keep the chart response compatible with the existing Custom Metrics chart format
- Preserve date range filtering and daily aggregation
- Exclude test configurations from the chart in REC/PROD

### Why
The previous chart used flow transition stats. In some cases, especially with RAG dialogs, a single user message can create multiple flow transition stat entries, which makes the chart display more messages than actually exist.
Counting dialog user actions directly avoids this overcount and aligns the chart with the “Total user messages” stats.

### DIFF
--- a/bot/admin/server/src/main/kotlin/BotAdminService.kt
+++ b/bot/admin/server/src/main/kotlin/BotAdminService.kt
@@ -47,6 +47,7 @@ import ai.tock.bot.admin.bot.sentencegeneration.BotSentenceGenerationConfigurati
 import ai.tock.bot.admin.bot.vectorstore.BotVectorStoreConfigurationDAO
 import ai.tock.bot.admin.dataset.DatasetDAO
 import ai.tock.bot.admin.dialog.ApplicationDialogFlowData
+import ai.tock.bot.admin.dialog.CountByDateResult
 import ai.tock.bot.admin.dialog.CountResult
 import ai.tock.bot.admin.dialog.DialogReport
 import ai.tock.bot.admin.dialog.DialogReportDAO
@@ -619,10 +620,20 @@ object BotAdminService {
         return grouped
     }
 
+    private fun groupCountByDateByAppConfigType(results: List<CountByDateResult>): Map<String, List<CountByDateResult>> {
+        val grouped: Map<String, List<CountByDateResult>> =
+            results.groupBy { stat ->
+                // At the moment, we rely on the `test-` prefix to distinguish test configurations
+                if (stat.applicationId.startsWith("test-")) APP_CONFIG_TEST_TYPE else APP_CONFIG_PROD_TYPE
+            }
+        return grouped
+    }
+
     fun getDialogStats(query: DialogStatsQuery): DialogStatsGroupResponse {
         val stats = dialogReportDAO.calculateDialogStats(query)
 
         val allUserActionsGroup = groupByAppConfigType(stats.allUserActions)
+        val allUserActionsByDateGroup = groupCountByDateByAppConfigType(stats.allUserActionsByDate)
         val allUserActionsExceptRagGroup = groupByAppConfigType(stats.allUserActionsExceptRag)
         val allUserRagActionsGroup = groupByAppConfigType(stats.allUserRagActions)
         val knownIntentUserActionsGroup = groupByAppConfigType(stats.knownIntentUserActions)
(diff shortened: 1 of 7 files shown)


bot/admin/web/src/app/metrics/metrics-board/metrics-board.component.ts
```diff
@@ -371,19 +368,16 @@ export class MetricsBoardComponent implements OnInit, OnDestroy {
...
-        !environment.production // In dev, we ask for test messages to dispose of some usable content
-      )
-    );
+    const url = `/dialogs/stats/messages-by-date`;
+    const payload = {
+      namespace: this.stateService.currentApplication.namespace,
+      applicationName: this.stateService.currentApplication.name,
+      from: this.range().start,
+      to: this.range().end,
+      includeTestConfigurations: !environment.production && !!this.displayTests
```
assouktim: We only include the tests if requested; you can remove `!environment.production`

bot/admin/server/src/main/kotlin/BotAdminService.kt
```diff
@@ -649,6 +652,36 @@ object BotAdminService {
...
+            dialogReportDAO
+                .countUserActionsByDate(query)
+                .groupBy { it.date }
+                .mapValues { (_, counts) -> counts.sumOf { it.total }.toInt() }
+        val dates = getDatesBetween(from.toLocalDate(), to.toLocalDate())
+
+        return UserAnalyticsQueryResult(
+            dates = dates,
+            usersData = dates.map { listOf(countsByDate[it] ?: 0) },
+            connectorsType = listOf("Messages"),
```
assouktim: I don't understand this line

bot/admin/server/src/main/kotlin/BotAdminService.kt
```diff
@@ -649,6 +652,36 @@ object BotAdminService {
...
+        )
+    }
```
assouktim: When I click “Display tests,” the graph showing the number of messages does not update.

bot/engine/src/main/kotlin/admin/dialog/DialogStatsQuery.kt
```diff
@@ -26,6 +26,7 @@ data class DialogStatsQuery(
     val applicationName: String,
     val from: ZonedDateTime? = null,
     val to: ZonedDateTime? = null,
+    val includeTestConfigurations: Boolean = false,
```
assouktim: Do not add `includeTestConfigurations`. We should follow the same approach as with the other metrics: we should get the statistics for production activity and those for the test separately.

bot/storage-mongo/src/main/kotlin/UserTimelineMongoDAO.kt
```diff
@@ -1163,6 +1165,25 @@ internal object UserTimelineMongoDAO : UserTimelineDAO, UserReportDAO, DialogRep
         return runBlocking { buildAndExecuteCountPipeline(namespace, applicationIds, filters) }
     }
 
+    override fun countUserActionsByDate(query: DialogStatsQuery): List<CountByDateResult> {
```
assouktim: Here's an alternative approach: Right now, the data you need is already being retrieved, but it's not grouped by date. You just need to add date grouping to the output.

assouktim (approved):
Here's an alternative approach: Right now, the data you need is already being retrieved, but it's not grouped by date. You just need to add date grouping to the output.

scezen:
Addressed the review points:

- Removed the custom messages endpoint/query
- Reused the existing dialog stats flow instead
- Kept the usual prod/test split and frontend merge behavior
- Fixed the refresh when toggling “Display tests”
- Fixed the 3D range display (it showed only last 2 days before that)

assouktim (approved):
Thanks!

rkuffer (approved):

PR #2049, opened 2026-05-07
[DERCBOT-1889] Refactor RAG Debug + DatasetRunStats Enhancement

scezen:


bot/admin/server/src/main/kotlin/service/DatasetService.kt
```diff
@@ -59,6 +59,17 @@ import java.time.Instant
 import java.util.Locale
 
 object DatasetService {
+    private val ragAnswerStatuses =
```
assouktim: It's better to create an enumeration for that.

bot/engine/src/main/kotlin/engine/action/SendSentenceWithFootnotes.kt
```diff
@@ -38,6 +38,8 @@ open class SendSentenceWithFootnotes(
     override var annotation: BotAnnotation? = null,
 ) :
     Action(playerId, recipientId, applicationId, id, date, state, metadata) {
+    var ragDebug: Any? = null
```
assouktim: Add `var ragDebug: Any? = null` after `override var annotation: BotAnnotation? = null` and then in the primary constructor. No need for a new constructor.

assouktim (changes requested):
A few changes to do

assouktim (approved):
Thanks

rkuffer (approved):

PR #2055, opened 2026-05-29
[DERCBOT-1924] Fix race condition when starting dataset run

scezen:
Ticket : [DERCBOT-1924](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1924)

  ### Goal
  Fix an issue where a dataset run could be marked as completed while some questions were never executed.

  ### Details
  - Create all dataset question tracking entries before making the run available to the worker
  - Prevent the worker from picking a run before all questions are prepared
  - Add a fallback in the worker when a question tracking entry is unexpectedly missing
  - Mark missing question results as failed with an explicit error instead of silently skipping them


assouktim (approved):
Thanks

rkuffer (approved):

PR #2056, opened 2026-05-29
[DERCBOT-1914] Enrich RAG context with source metadata and chunk ids

scezen:
Ticket: [DERCBOT-1914](https://jiradc.intra.arkea.com:8443/jira/browse/DERCBOT-1914)

### Goal
Expose richer RAG context metadata so the LLM can identify the exact source chunk and return web page URLs when available.

### Details
- Add `title` and `source_url` to the JSON context injected into the RAG prompt
- Populate `source_url` only when the document source is an HTTP(S) URL
- Change `chunk_id` to concatenate the document id and chunk value
- Match RAG footnotes using the new composite chunk identifier
- Update the default RAG prompt instructions to require the exact `chunk_id`

### DIFF
--- a/bot/admin/web/src/app/rag/rag-settings/models/engines-configurations.ts
+++ b/bot/admin/web/src/app/rag/rag-settings/models/engines-configurations.ts
@@ -347,7 +347,7 @@ Must list ALL retrieved chunks.
 
 For each chunk:
 
-- chunk: identifier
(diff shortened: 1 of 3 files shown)


assouktim (approved):
Thanks!

rkuffer (approved):

PR #2057, opened 2026-06-02
[DERCBOT-1649][RAG v3] Multi-Query Retriever

assouktim:


gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/rag_chain_builder.py
```diff
@@ -0,0 +1,440 @@
...
+        hybrid_retriever = HybridRetriever(
+            vector_retriever=vector_retriever,
+            fts_retriever=fts_retriever,
+            rrf_top_n=request.document_search_params.k,
+        )
+
+        retriever = RunnableLambda(
+            name="hybrid_retrieve", func=hybrid_retriever.retrieve
+        )
+
```
scezen, line 409: Looks like "compressor_setting" is not used anymore, is it intentional?
The retriever is now created directly, so I think existing BloomzRerank configs are not applied anymore?

scezen (approved):

PR #2062, opened 2026-07-06
[FRONT] DERCBOT-1981 fix: prevent lexicon group id 0 on empty base

rkuffer:
When the backend returns an empty lexiconGroups list on component init, the front-side id sequence (_lexIdSeq) falls back to 0. The first lexicon group created then inherits id 0 (post-increment), which the backend rejects on save with Lexicon group ids must be positive.
Minimal front-side fix: initialize the _lexIdSeq fallback to 1 instead of 0 in prompt-context-settings, so the first group created on an empty base gets a positive id.
Note: the group id is a purely front-end tracking key but is still sent to the backend. A follow-up fix (dropping this id from the save payload) should be considered to remove the coupling entirely.


assouktim (approved):
Thanks

scezen (approved):

PR #2065, opened 2026-07-16
[DERCBOT-1986] Support custom sources label

assouktim:
The Google Chat connector currently displays a fixed "Sources" label when adding source references at the end of responses.

This change makes the sources label configurable, allowing the displayed text to be customized depending on the use case or integration requirements.


bot/connector-google-chat/src/main/kotlin/GoogleChatConnectorProvider.kt
```diff
@@ -108,6 +109,9 @@ internal object GoogleChatConnectorProvider : ConnectorProvider {
             val useThread =
                 connectorConfiguration.parameters[USE_THREAD_PARAMETER] == "1"
 
+            val sourcesLabel =
+                connectorConfiguration.parameters[SOURCES_LABEL_PARAMETER] ?: "Sources"
```
scezen: If a user fills in this field and then clears it, the UI may save an empty string. In that case, "Sources" will not be used and the source title will be empty. I think we should handle blank values here, like we do for introMessage above?

assouktim: absolutely

scezen (approved):

scezen (approved):

rkuffer (approved):

PR #2068, opened 2026-07-29
fix #2066 [Web Connector] add check for updated frontend locale

### LINKED ISSUE #2066: [Web Connector] changing the language in the frontend does not update user preferences
Using a frontend like `tock-react-kit`, we have the option to set the user locale for dialogs with TOCK chatbots. However, when using the `tock_web_connector_persist_profile` option, past the session initialization, changing the value in the frontend/in network calls does not lead to any change in the backend.

Fabilin:
fixes #2066 by adding a new method to the `Connector` interface, which allows more fine-grained control over when `UserPreferences` are refreshed. The default implementation keeps the old behavior.


zigzago:
@Fabilin ping me when the PR is ready

zigzago (approved):

PR #2074, opened 2026-08-13
feat(studio): add configurable environment banner

rkuffer:
## ⚠️ Merge order

**Do not merge before #2073 (chore/studio-angular-21).** This branch is stacked
on top of it and depends on the Angular 21 upgrade (notably
`provideAppInitializer`, introduced in Angular 19+).

## What

Adds an optional banner to the Studio indicating the current environment
(e.g. `REC`, `STAGING`), so users can tell at a glance which instance they
are working on. Nothing is shown in production by default.

The banner is made of two parts:
- a **diagonal ribbon** in the bottom-left corner, dismissable per browser tab;
- a permanent **4px colored strip** along the top of the window, which cannot
  be dismissed — so the environment stays identifiable even after the ribbon
  is hidden.

Placement was chosen to avoid overlapping existing UI: the header actions
(top-right), the back-to-top button (bottom-right), and the main menu (top and
left) are all left untouched.

## How

The banner is driven by a single runtime file, `src/assets/env-banner.json`,
read once at application startup via an app initializer. A **neutral** version
(`{ "label": null }`) is committed to the repository, so no `404` is emitted
when no banner is configured.

Resolution order: runtime file > build-time `environment.envBanner` default >
nothing. An empty or missing label resolves to no banner, meaning a
misconfiguration can never accidentally display a banner in production.

The Studio ships as a single, environment-agnostic build: the banner is
resolved **at runtime, not at build time**, so the environment-specific file
is injected at deployment by overwriting the served asset. This preserves
image promotion — the same artifact moves unchanged from one environment to
the next.

For local testing, a label can be forced via `?env-banner=REC` without
touching the file.

## Config

| Field       | Required | Description                                             |
| ----------- | -------- | ------------------------------------------------------- |
| `label`     | yes      | Text shown in the banner. `null` or absent = no banner. |
| `color`     | no       | Background color. Defaults to amber.                    |
| `textColor` | no       | Text color. Auto-derived from `color` when omitted.     |

Full documentation, including per-environment CI/deployment patterns
(Kubernetes ConfigMap, container entrypoint), is added to the Studio README.

## Notes

- New standalone component (`EnvBannerComponent`) + resolver service, mounted
  as a sibling of the layout in the root component.
- i18n keys added for the dismiss control (Transloco).
- No backend change, no new endpoint.

### DIFF
--- a/bot/admin/web/README.md
+++ b/bot/admin/web/README.md
@@ -37,6 +37,102 @@ To preview what the hook would replace without modifying any file:
 DRY_RUN=1 bash .git/hooks/pre-commit
 ```
 
+## Environment banner
+
+The Studio can display a banner indicating the current environment (e.g. `REC`,
+`STAGING`), so users can tell at a glance which instance they are working on. It
+is driven by a single runtime file, `src/assets/env-banner.json`, read once at
+application startup.
+
+A **neutral** version of this file is committed to the repository:
+
+```json
+{ "label": null, "_comment": "Set label to display an environment banner, e.g. 'REC'. Optional: color, textColor." }
+```
+
+With `label` unset, no banner is shown. This is the intended state for
+production and for any instance that should look "clean". Because the file is
+present and valid, no `404` request is emitted at startup.
+
+### Displaying a banner
+
+To show a banner on a given environment, serve an `env-banner.json` with a
+non-empty `label`:
+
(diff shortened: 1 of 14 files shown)


assouktim (approved):
Thanks!

scezen (approved):

PR #2075, opened 2026-08-13
feat(studio): add CSV import/export for datasets

rkuffer:
## ⚠️ Merge order
Do not merge before https://github.com/theopenconversationkit/tock/pull/2073 (chore/studio-angular-21). This branch is stacked on top of it and depends on its i18n implementation

## What

Extends the existing JSON import/export mechanism with CSV support, using the papaparse dependency already present in the studio.

## Changes

Export (dataset-create): the export button in edit mode now offers both JSON and CSV. JSON exports the full dataset; CSV exports questions only (question, groundTruth), since name and description have no place in a flat two-column file.

Import (datasets-board): the import dialog accepts .json and .csv. Because a CSV carries no name/description, those become required inputs shown only for CSV imports (name is pre-filled from the filename). A "download CSV template" button provides an empty file with the expected columns.

New dataset-csv.ts: pure, framework-free module wrapping papaparse — questionsToCsv, buildCsvTemplate, csvToQuestions, and a typed CsvParseError. Handles delimiter auto-detection on import (; , tab), UTF-8 BOM + ; on export for Excel FR compatibility, header normalization, and CSV-injection sanitization on export.

FileValidators.extensionSupported: new validator matching on file extension rather than MIME type. Aligns file validation with the upload widget's fileTypeAccepted (which already filters by extension) and avoids the unreliable CSV MIME reported across OSes.

## Notes
Fixed a latent typing issue surfaced by this work: the board's import path previously fed JSON.parse() output (typed any) to createDataset, silently bypassing type checking. The import payload type is now derived from the service signature.
Design choice: fixed CSV structure (question;groundTruth) rather than user-configurable column mapping — deliberate, to avoid overhead disproportionate to the need. The parser is structured to allow custom mapping later if required.
Client-side validators remain UX guards; actual content validation stays in csvToQuestions / JSON.parse and the backend.

### DIFF
--- a/bot/admin/web/package-lock.json
+++ b/bot/admin/web/package-lock.json
@@ -78,6 +78,7 @@
         "@types/jasminewd2": "^2.0.8",
         "@types/katex": "^0.16.7",
         "@types/node": "^20.0.0",
+        "@types/papaparse": "^5.5.2",
         "@typescript-eslint/eslint-plugin": "^8.66.0",
         "@typescript-eslint/parser": "^8.66.0",
         "eslint": "^8.57.0",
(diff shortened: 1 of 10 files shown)


assouktim (approved):
Thanks!

scezen (approved):

scezen (approved):

PR #2077, opened 2026-08-19
resolves #2076 [Web Connector] add props to routing context

### LINKED ISSUE #2076: [Web Connector] additional context for `WebSecurityHandler` implementations
When implementing a custom `WebSecurityHandler`, there is currently no practical way to know the connector on which the request happens, nor the path on which the browser calls the connector. This is already limiting for the built-in `WebSecurityCookiesHandler` implementation, which cannot set a proper path for its cookie and has to rely on the default browser handling. Custom implementations, notably OAuth ones, are further impacted, as they need to store large and secured cookies, and may need to distinguish between several web connectors.

This context could additionally be useful to any third-party handler tacked on a web connector, meaning it would be best to make its storage and retrieval quite generic.

Fabilin:
resolves #2076 by storing the connector ID and public path in the routing context for any web connector request. "Public path" is defined as a separate variable from the existing "path", in order to support reverse proxies.


Fabilin:
@zigzago CI passes now 👍

zigzago (approved):

PR #2080, opened 2026-08-20
[DERCBOT-1913] Allow FAQ Import

scezen:
### Goal
Allow FAQs exported from one environment to be imported into another application.

### Details
- Add a backend endpoint to import the FAQ file
- Check the imported data before saving it
- Create new FAQs or update existing ones in the target application
- Add an import button and dialog to the FAQ management page
- Refresh the FAQ list after a successful import


bot/admin/server/src/main/kotlin/FaqAdminService.kt
```diff
@@ -142,6 +143,108 @@ object FaqAdminService {
...
+            badRequest("The FAQ import is empty")
+        }
+
+        val duplicatedIntentNames =
+            queries.groupingBy { it.intentName }.eachCount().filterValues { it > 1 }.keys
+        if (duplicatedIntentNames.isNotEmpty()) {
+            badRequest("Duplicated FAQ intent names in import: ${duplicatedIntentNames.joinToString()}")
+        }
+
+        val queriesWithExistingFaq =
```
assouktim, line 192: **Simplify the validation**

This check is redundant: it first looks up the FAQ, then the story. A FAQ is just a story with the faq category, so checking the story alone is enough. The rule per intent becomes:

- no existing story, or existing story with FAQ category → create / update;
- existing story with another category → error.

The FaqDefinition lookup shouldn't happen here at all: it's only needed when the import actually proceeds. Move it to the processing step that follows validation, so we don't hit the database for data we'll throw away if the import fails.

**Error reporting**

Rather than failing fast on the first invalid intent (badRequest inside the map), I'd validate all intents, collect the failing ones, and only then fail the import with a single explicit message, e.g.:

_"Import failed: the following FAQs do not correspond to FAQs in the bot — [list of intents]. They exist as stories of a different type and cannot be updated. Please first review whether these non-FAQ stories are still relevant and delete them if possible; otherwise, remove the listed FAQs from the import."_

This way the user sees all conflicts in one pass instead of discovering them one at a time.

bot/admin/server/src/main/kotlin/FaqAdminService.kt
```diff
@@ -142,6 +143,108 @@ object FaqAdminService {
...
+                ),
+                userLogin,
+                application,
+            )
+        }
+
+        return queriesWithExistingFaq.size
+    }
+
+    private fun validateImportedFAQ(
```
assouktim, line 245: Similarly, this should be added to the list of errors. Do not throw the exception here.

assouktim (changes requested):
Thank you, there are just two improvements to be made.

bot/admin/web/src/app/faq/faq.module.ts
```diff
@@ -48,6 +48,7 @@ import { FaqManagementListComponent } from './faq-management/faq-management-list
 import { FaqManagementEditComponent } from './faq-management/faq-management-edit/faq-management-edit.component';
 import { BotSharedModule } from '../shared/bot-shared.module';
 import { FaqManagementSettingsComponent } from './faq-management/faq-management-settings/faq-management-settings.component';
+import { FaqManagementImportComponent } from './faq-management/faq-management-import/faq-management-import.component';
```
assouktim, line 51: Missing file: faq-management-import.component

The build is failing because faq.module.ts imports FaqManagementImportComponent but the component file doesn't exist (or wasn't committed). Please add the missing file or fix the import path.

`Error TS2307: Cannot find module './faq-management/faq-management-import/faq-management-import.component'`

assouktim (changes requested):

assouktim (approved):
Looks Good To Me. Thanks

PR #2087, opened 2026-08-27
[DERCBOT-1686] Google Chat Feedback feature

scezen:
### Goal

Allow users to rate Google Chat bot answers with a thumbs-up or thumbs-down button, without polling or Pub/Sub subscriptions.

### Details

- Add optional thumbs-up and thumbs-down accessory buttons to Google Chat bot answers
- Handle button click events directly through the existing connector endpoint
- Store the selected vote using Tock's existing action feedback mechanism
- Replace both feedback buttons with the selected option in a disabled state after the first vote
- Prevent later clicks from replacing an existing Google Chat feedback
- Keep feedback disabled by default to preserve the current connector behavior
- Exclude waiting and introductory messages from feedback
- Add connector configuration, documentation, and unit tests


assouktim (approved):
Thanks

rkuffer (approved):

PR #2095, opened 2026-09-14
[FRONT FIX] Add Angular dev proxy config for Studio front-to-back communication

rkuffer:
What

Adds a proxy.conf.js to the Studio web app and wires it into the serve target in angular.json, so ng serve proxies /rest calls to the local BotAdmin server (same-origin). The dev environment.ts now uses a relative serverUrl (/rest) instead of an absolute one.

Why

Running the Studio front in dev (ng serve) against a BotAdmin started from sources currently doesn't work out of the box: the authentication request never completes and the app stays stuck on startup. The project has never had a dev proxy config — the front was wired to call the back via an absolute URL, a setup dating back to the project's Angular 2 origins. A proxy.conf referenced by the serve target is the standard way to run an Angular app against its backend in dev today. This change makes the local dev setup work without per-developer workarounds, and has no effect on production builds (serve is dev-only, build is unaffected).

Note on the Accept-Encoding: identity header

The proxy sets Accept-Encoding: identity. This is a workaround, not a preference, and it's commented as such in the file. During investigation we found that BotAdmin hangs on any request that negotiates gzip: the server returns a compressed response the client never manages to finalize, so the request stays pending indefinitely. This is reproducible with plain curl, independently of the front-end:

```
# hangs:
curl -i -H 'Accept-Encoding: gzip' -X POST http://localhost:<port>/rest/authenticate \
  -H 'Content-Type: application/json' -d '{"email":"...","password":"..."}'
# responds normally:
curl -i -H 'Accept-Encoding: identity' -X POST http://localhost:<port>/rest/authenticate ...
```

Forcing identity avoids the broken compression path. We narrowed this down to a server-side issue (present at least up to Vert.x 5.0.12; the 5.0.12→5.1.3 bump is not the cause). We could not conclusively determine whether it's a regression introduced with the Vert.x 5 upgrade or a pre-existing latent behaviour, as building older revisions ran into unrelated toolchain issues. This likely deserves a separate server-side issue; the header can be removed here once the underlying compression behaviour is fixed.


assouktim (approved):
Thanks

scezen (approved):

